PR #1, opened 2017-08-03
Add a logging system

rkrzr:
Using putStrLn in a number of concurrent threads leads to garbled output.
We therefore use a TBQueue for log messages as well and have a separate
logging loop that writes them to stdout one after the other leading to
nice logs.

### DIFF
--- a/src/Core.hs
+++ b/src/Core.hs
@@ -9,6 +9,7 @@ module Core
   getCurrentValue,
   handleOp,
   lookup,
+  log,
   newCore,
   postQuit,
   processOps
@@ -21,8 +22,9 @@ import Control.Concurrent.STM.TBQueue (TBQueue, newTBQueueIO, readTBQueue, write
 import Control.Concurrent.STM.TVar (TVar, newTVarIO, writeTVar, readTVar)
 import Control.Monad (unless)
 import Data.Aeson (Value (..))
+import Data.Text (Text)
 import Data.UUID (UUID)
-
+import Prelude hiding (log)
 import Store (Path)
 import Subscription (SubscriptionTree, empty)
 
@@ -46,18 +48,25 @@ data Updated = Updated Path Value deriving (Eq, Show)
 
 data EnqueueResult = Enqueued | Dropped
 
+type LogRecord = Text
+
 data Core = Core
   { coreCurrentValue :: TVar Value
   , coreQueue :: TBQueue (Maybe Op)
   , coreUpdates :: TBQueue (Maybe Updated)
   , coreClients :: MVar ServerState
+  , coreLogRecords :: TBQueue (Maybe LogRecord)
   }
 
 type ServerState = SubscriptionTree UUID WS.Connection
 
+log :: LogRecord -> Core -> IO ()
+log record core = atomically $ do
+  isFull <- isFullTBQueue (coreLogRecords core)
+  unless isFull $ writeTBQueue (coreLogRecords core) (Just record)
+
 newServerState :: ServerState
 newServerState = empty
-
 
 newCore :: IO Core
 newCore = do
@@ -65,7 +74,8 @@ newCore = do
   tqueue <- newTBQueueIO 256
   tupdates <- newTBQueueIO 256
   tclients <- newMVar newServerState
-  pure (Core tvalue tqueue tupdates tclients)
+  tlogrecords <- newTBQueueIO 256
+  pure (Core tvalue tqueue tupdates tclients tlogrecords)
 
 -- Tell the put handler loop and the update handler loop to quit.
 postQuit :: Core -> IO ()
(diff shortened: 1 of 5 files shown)


ruuda:
Why not use `monad-logger` for this? It has the additional advantage that it can be set to a noop logger in tests. (Whereas with a bounded queue, it would fill up and block, and an unbounded queue is a bad idea.)

rightfold:
I generally avoid message processing loops if a mutex would do.

rkrzr:
> I generally avoid message processing loops if a mutex would do.

The mutex could be contended though if many threads are trying to write at the same time, right? Using a queue should be better throughput-wise, I would think.

src/WebsocketServer.hs
```diff
@@ -57,15 +60,16 @@ handleClient conn path core = do
   uuid <- newUUID
   let
     state = coreClients core
+    logRecords = coreLogRecords core
     onConnect = do
       modifyMVar_ state (pure . Subscription.subscribe path uuid conn)
-      putStrLn $ "Client " ++ (show uuid) ++ " connected, subscribed to " ++ (show path) ++ "."
+      log (T.pack $ "Client " ++ (show uuid) ++ " connected, subscribed to " ++ (show path) ++ ".") logRecords
```
NunoAlexandre: Is `T.pack` necessary? I'd expect the `OverloadedStrings` extension to take care of this.

src/WebsocketServer.hs
```diff
@@ -77,9 +81,10 @@ handleClient conn path core = do
 -- closed. So we go into an infinite loop here.
 keepTalking :: WS.Connection -> IO ()
 keepTalking conn = forever $ do
-    -- Note: WS.reciveDataMessage will handle control messages automatically and e.g.
+    -- Note: WS.receiveDataMessage will handle control messages automatically and e.g.
     -- do the closing handshake of the websocket protocol correctly
-    WS.receiveDataMessage conn
+    _ <- WS.receiveDataMessage conn
+    putStrLn "bla"
```
NunoAlexandre: Is `bla` intended?

src/WebsocketServer.hs
```diff
@@ -36,15 +39,15 @@ broadcast =
   let
     send :: Value -> WS.Connection -> IO ()
     send value conn = do
-      putStrLn $ "Sending text data ..."
+      -- putStrLn $ "Sending text data ..."
```
NunoAlexandre: Delete line?

src/Core.hs
```diff
@@ -97,11 +105,10 @@ processOps core = go Null
...
-          atomically $ writeTVar (coreCurrentValue core) newValue
-          atomically $ writeTBQueue (coreUpdates core) (Just $ Updated (opPath op) newValue)
+          atomically $ do
+            writeTVar (coreCurrentValue core) newValue
+            writeTBQueue (coreUpdates core) (Just $ Updated (opPath op) newValue)
           go newValue
         Nothing -> do
-          -- Stop the loop when we receive a Nothing. Tell the update loop to
-          -- quit as well.
-          atomically $ writeTBQueue (coreUpdates core) Nothing
```
NunoAlexandre: Did you mean to delete this line? How will it signal a `stop` in the update and log queues? Or, how was this useful before and no longer is?

src/WebsocketServer.hs
rightfold: `OverloadedStrings` only works with string literals. You can concatenate `Text` string literals with `(<>)` from `Data.Semigroup`, instead of `(++)` which is exactly `(<>)` but specific to `[a]`. Also you'd have to `T.pack` the `show`s anyway.

NunoAlexandre:
`monad-logger` seems indeed like a good solution and less invasive. I would merge this PR and look into that after.

> I generally avoid message processing loops if a mutex would do

I'd rather stick to STM than go for mutexes. I never used mutexes in Haskell but in C. It's not pretty.

src/WebsocketServer.hs
rkrzr: Removed.

src/WebsocketServer.hs
rkrzr: Yeah, because we anyway use `show` it's necessary here.

src/WebsocketServer.hs
rkrzr: Removed.

src/Core.hs
rkrzr: This was moved up to `postQuit`, which is a more logical place to put this, I'd say.

rkrzr:
@ruuda `monad-logger` would perhaps be nicer, but I would have to figure out how to use it for multiple threads. I'll merge this for now and we can always improve upon it later.

PR #3, opened 2017-08-31
Add a simple Websocket client example site

rkrzr:
Add a simple JS example listener.

### DIFF
--- a/integration-tests/listener.html
+++ b/integration-tests/listener.html
@@ -1,6 +1,12 @@
 <!DOCTYPE html>
 <meta charset="utf-8"/>
 <title>A simple Websocket listener</title>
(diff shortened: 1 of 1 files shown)


ruuda:
Helvetica or not, I think an opening `<html>` tag might be appropriate.

arianvp:
Nope it's not required by the html5 spec

ruuda:
TIL. A `<div>` inside a `<p>` is not allowed though. You can use a `<span>` instead.

rkrzr:
Thanks for correcting my HTML and style violations. That Helvetica really makes all the difference @duijf !
I'll try to use more JS dependencies next time.

PR #6, opened 2017-09-06
Add Haskell client library

rightfold:
Now we can write the integration tests in Haskell. :+1:


server/integration-tests/listener.html
```diff
@@ -17,7 +17,6 @@
     console.log('Received data: ' + event.data);
     var status = document.getElementById("status");
     console.log('status: ' + status);
-    status.innerHTML = JSON.stringify(msg);
```
rkrzr: Good catch. This is what happens when I try to write some Javascript.

README.md
```diff
@@ -16,12 +16,13 @@ Do not use it.
...
 Run the tests with `stack test`.
 And run `icepeak` itself with `stack exec icepeak`.
-Install with `stack install`
+Install with `stack install`.
 
-Integration tests are in `/integration-tests`.
+Integration tests are in `/server/integration-tests`.
 They are stand-alone scripts that can be executed directly, e.g. `./connection_test.py`.
 
 ## Usage:
```
rkrzr: Can you add some instructions for the client as well?

client-haskell/src/Icepeak/Client.hs
```diff
@@ -0,0 +1,66 @@
...
+  , clientIcepeakPort :: Word16
+  , clientIcepeakAuth :: ByteString
+  }
+
+-- | Set a value at the leaf of a path.
+--
+-- Will rethrow any exceptions thrown by the I/O actions from
+-- "Network.HTTP.Client". Will not throw any other exceptions.
+setAtLeaf :: (MonadIO m, ToJSON a) => Client -> [Text] -> a -> m HTTP.Status
+setAtLeaf (Client http host port auth) path leaf =
```
rkrzr: s/http/httpManager

client-haskell/src/Icepeak/Client.hs
```diff
@@ -0,0 +1,66 @@
...
+    { HTTP.method = "PUT"
+    , HTTP.path = requestPathForIcepeakPath path
+    , HTTP.requestBody = HTTP.RequestBodyLBS (Aeson.encode leaf)
+    }
+
+-- | Return the request path for an Icepeak path.
+requestPathForIcepeakPath :: [Text] -> ByteString
+requestPathForIcepeakPath =
+  ByteString.Lazy.toStrict .
+    Binary.Builder.toLazyByteString .
```
rkrzr: Is there no way to directly get a strict bytestring from the path?

client-haskell/src/Icepeak/Client.hs
```diff
@@ -0,0 +1,66 @@
...
+
+-- | Set a value at the leaf of a path.
+--
+-- Will rethrow any exceptions thrown by the I/O actions from
+-- "Network.HTTP.Client". Will not throw any other exceptions.
+setAtLeaf :: (MonadIO m, ToJSON a) => Client -> [Text] -> a -> m HTTP.Status
+setAtLeaf (Client http host port auth) path leaf =
+  let request = setAtLeafRequest path leaf
+                & (\r -> r { HTTP.host = host, HTTP.port = fromIntegral port })
+                & HTTP.setQueryString [("auth", Just auth)]
```
rkrzr: Couldn't you set the `host`, the `port` and the query string in the `setAtLeafRequest` function?
Then all the request building code would be in one place.

client-haskell/src/Icepeak/Client.hs
rightfold: Nope. :sob:

rightfold:
> Perhaps you can still add a Haskell integration test that demonstrates how to use the client?

PR #7, opened 2017-09-15
Add delete functionality to Haskell client library

rightfold:


client-haskell/src/Icepeak/Client.hs
```diff
@@ -37,28 +48,49 @@ data Client = Client
...
 -- "Network.HTTP.Client". Will not throw any other exceptions.
+
+-- | Set a value at the leaf of a path.
 setAtLeaf :: (MonadIO m, ToJSON a) => HTTP.Manager -> Client -> [Text] -> a -> m HTTP.Status
 setAtLeaf httpManager client path leaf =
   let request = setAtLeafRequest client path leaf
   in liftIO . fmap HTTP.responseStatus $ HTTP.httpNoBody request httpManager
 
+-- | Delete the value at the leaf of a path.
+deleteAtLeaf :: MonadIO m => HTTP.Manager -> Client -> [Text] -> m HTTP.Status
```
rkrzr: Is there a concrete advantage of having this be an `MondadIO m` instead of just being `IO`?
Does servant roll its own IO monad or so?

rightfold: So that the caller does not need to use `liftIO`.

PR #8, opened 2017-09-19
Notify subscribers of changes to child nodes

### LINKED ISSUE #5: Icepeak does not push non-root updates to the WebSocket clients
For example, `PUT /foo/bar` does not cause the client to receive the update. Only `PUT /` does.

rightfold:
Fixes #5.

Also adds more functionality to `listener.html`, that is useful for debugging.
Also adds a pure variant of `broadcast`, which returns a list of notifications instead of sending the notifications.

### DIFF
--- a/server/integration-tests/listener.html
+++ b/server/integration-tests/listener.html
@@ -5,18 +5,41 @@
 <style>
   body { font-family: 'Helvetica', 'Arial', sans-serif; }
 </style>
-<p>
+<fieldset>
+  <legend>Configuration</legend>
+  URL: <input id="url" type="url" value="ws://localhost:3000"/>
+  <button id="reload" type="button">Reload</button>
+</fieldset>
+<fieldset>
+  <legend>Status</legend>
   <!-- we set the status dynamically from the websocket data -->
-  Status: <span id="status"></span>
-</p>
+  <pre id="status"></pre>
+</fieldset>
 
 <script>
-  var webSocket = new WebSocket("ws://localhost:3000");
-  webSocket.onmessage = function (event) {
-    var msg = JSON.parse(event.data);
-    console.log('Received data: ' + event.data);
-    var status = document.getElementById("status");
-    console.log('status: ' + status);
-    status.textContent = JSON.stringify(msg);
-  }
+  (function() {
+    var urlField = document.getElementById("url");
+    var reloadButton = document.getElementById("reload");
+    var statusField = document.getElementById("status");
+
+    var webSocket = null;
+    var reload = function() {
+      if (webSocket !== null) {
+        webSocket.close();
+      }
+      webSocket = new WebSocket(urlField.value);
+      webSocket.addEventListener("message", function (event) {
+        var msg = JSON.parse(event.data);
+        console.log('Received data: ' + event.data);
+        console.log('status: ' + status);
+        statusField.textContent = JSON.stringify(msg, null, 4);
+      });
+    };
+
+    reloadButton.addEventListener("click", function() {
+      reload();
+    });
+
+    reload();
+  })();
 </script>
--- a/server/src/Subscription.hs
+++ b/server/src/Subscription.hs
@@ -19,13 +19,13 @@ import Data.Aeson (Value)
 import Data.Foldable (traverse_)
 import Data.HashMap.Strict (HashMap)
 import Data.Hashable (Hashable)
-import Data.List (inits)
 import Data.Maybe (fromMaybe)
 import Data.Monoid ((<>))
 import Data.Text (Text)
 
 import qualified Control.Concurrent.Async as Async
 import qualified Data.HashMap.Strict as HashMap
+import qualified Data.List.NonEmpty as NE
 import qualified Data.Text as Text
 
 import qualified Store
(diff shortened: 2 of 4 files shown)


server/integration-tests/listener.html
```diff
@@ -5,18 +5,41 @@
 <style>
   body { font-family: 'Helvetica', 'Arial', sans-serif; }
 </style>
-<p>
+<fieldset>
+  <legend>Configuration</legend>
+  URL: <input id="url" type="url" value="ws://localhost:3000"/>
+  <button id="reload" type="button">Reload</button>
+</fieldset>
+<fieldset>
```
rkrzr: Oh, fancy!

server/integration-tests/listener.html
```diff
@@ -5,18 +5,41 @@
...
+  <legend>Status</legend>
   <!-- we set the status dynamically from the websocket data -->
-  Status: <span id="status"></span>
-</p>
+  <pre id="status"></pre>
+</fieldset>
```
rkrzr: HTML really changed since 1999.

server/src/Subscription.hs
```diff
@@ -64,22 +68,30 @@ unsubscribe path subid (SubscriptionTree here inner) =
...
-broadcast :: (Value -> v -> IO ()) -> [Text] -> Value -> SubscriptionTree k v -> IO ()
-broadcast f path value (SubscriptionTree here inner) =
+broadcast :: (v -> Value -> IO ()) -> [Text] -> Value -> SubscriptionTree k v -> IO ()
+broadcast f path value tree =
+  let notifications = execWriter $ broadcast' path value tree
+  in Async.mapConcurrently_ (uncurry f) notifications
+
+-- Like broadcast, but return a list of notifications rather than invoking an
+-- effect on each of them.
+broadcast' :: [Text] -> Value -> SubscriptionTree k v -> Writer [(v, Value)] ()
```
rkrzr: Could you add some docs to explain the use of Writer here?
Is the main advantage that we have fewer values to pass around in recursive calls?

Maybe also mention that `v` is one websocket connection here. Or better just call it `connection` instead of `v`.
(It's nice to abstract this, for testing, but it makes it now also harder to follow along).

server/src/Subscription.hs
```diff
@@ -64,22 +68,30 @@ unsubscribe path subid (SubscriptionTree here inner) =
...
+broadcast' path value (SubscriptionTree here inner) = do
   case path of
     [] -> do
       -- When the path is empty, all subscribers that are "here" or at a deeper
       -- level should receive a notification.
       -- We broadcast concurrently since all updates are independent of each other
```
rkrzr: Please move this comment up to where `Async.mapConcurrently` is actually called.

server/src/Subscription.hs
```diff
@@ -64,22 +68,30 @@ unsubscribe path subid (SubscriptionTree here inner) =
...
-    key : pathTail -> case HashMap.lookup key inner of
-      Nothing -> pure ()
-      -- TODO: Extract the inner thing from the value as well; the client is not
-      -- subscribed to the top-level thing after all.
-      Just subs -> broadcast f pathTail (Store.lookupOrNull [key] value) subs
+    key : pathTail -> do
+      traverse_ (\v -> tell [(v, value)]) here
+      for_ (HashMap.lookup key inner) $ \subs ->
+        -- TODO: Extract the inner thing from the value as well; the client is not
+        -- subscribed to the top-level thing after all.
```
rkrzr: Isn't this fixed now? I thought that a client will now only receive the sub-tree that it is subscribed to and not the whole tree any more.

server/tests/SubscriptionTreeSpec.hs
```diff
@@ -69,3 +72,44 @@ spec = do
...
+                     , (conn2, value_foo)
+                     , (conn3, value_foo_bar)
+                     ]
+
+      it "notifies parents and children about updates" $ do
+        broadcast'' ["foo", "bar"]
+          `shouldBe` [ (conn1, value)
+                     , (conn2, value_foo)
+                     , (conn3, value_foo_bar)
+                     ]
```
rkrzr: Please add  a test that shows that a client subscribed to a subtree will *only* receive that subtree on a change and not the whole tree.

server/src/Subscription.hs
rightfold: Seems to have been already fixed here:

https://github.com/channable/icepeak/commit/77f8a90a59f616aa7b8631b109d4d57a866f3845

Will delete the comment.

server/integration-tests/listener.html
arianvp: `fieldset` and `pre` were added in December 1999 😄

server/tests/SubscriptionTreeSpec.hs
rightfold: The tests catch that; if the clients would receive the whole tree then the test will fail when checking equality with `value_foo` and `value_foo_bar`.

server/src/Subscription.hs
rightfold: Without it, recursion would be a bit trickier, we'd have to use `foldMap` and `(<>)`. With `Writer` we can use `do`-notation which is less obscure. I'd like to get rid of `Writer` from the type signature.

server/tests/SubscriptionTreeSpec.hs
rkrzr: Ok. Is there also a test that makes sure that a client does *not* get updated if an unrelated part of the tree is updated?

rightfold: Yes; `conn4` is not present in all of the test examples.

Issue #9, opened 2017-09-21
Add MVP authentication and authorization


rightfold:
The current situation is that requests must include a token which is set when starting Icepeak. It is not possible to customize which users may touch what data.

We will implement support for authorization by switching to [JWT][jwt]. The JWT token will include a claim called `"icepeak"` that has the following JSON schema:

```json
{
  "type": "object",
  "required": ["version"],
  "properties": {
    "version": {"enum": [1]},
    "whitelist": {
      "type": "array",
      "items": {
        "type": "object",
        "required": ["prefix", "modes"],
        "properties": {
          "prefix": {"type": "array", "items": {"type": "string"}},
          "modes": {"type": "array", "items": {"enum": ["read", "write"]}}
        }
      }
    }
  }
}
```

Example JWT token claim set:

```json
{
  "icepeak": {
    "version": 1,
    "whitelist": [
      {"prefix": ["foo"], "modes": ["read"]},
      {"prefix": ["bar", "1"], "modes": ["read", "write"]}
    ]
  }
}
```

A request with this claim set may only:
 - Read paths that have `/foo` as a prefix.
 - Update paths that have `/bar/1` as a prefix.
Any other request will result in a *401 Unauthorized* response.

A token can be specified in an `Authorization: Bearer <token>` header or an `access_token=<token>` query string parameter.

Requests that do not include a token will always be responded to with a *401 Unauthorized* error.

[jwt]: https://tools.ietf.org/html/rfc7519

fatho:
Should we also differentiate between read and write access?

rightfold:
Good point. I will add that to the schema.

rkrzr:
> Should we also differentiate between read and write access?

Yes, but let's support all of create, read, update and delete.

rightfold:
Added those to the schema.

fatho:
We need to decide on a semantics for whitelisting, in particular when there are multiple entries that apply to the same path, as in the following example. (@rkrzr: regarding our discussion yesterday)

Suppose we have the following whitelist

```
{"prefix": ["foo"], "modes": ["read"]},
{"prefix": ["foo"], "modes": ["update"]}
```

Then I'd say that this is valid and would allow both `read` and `update` actions on all paths prefixed by `/foo/`, because for all such actions there is an entry that allows it.

Similarly, the whitelist

```
{"prefix": ["foo"], "modes": ["update"]},
{"prefix": ["foo", "bar"], "modes": ["read"]}
```

[...]

PR #11, opened 2017-09-22
Add support for authentication and authorization through JSON Web Tokens

fatho:
For more information, see issue #9.


server/app/Icepeak.hs
```diff
@@ -51,7 +52,8 @@ readValue filePath = do
 
 main :: IO ()
 main = do
-  config <- execParser configInfo
+  env <- getEnvironment
+  config <- execParser (configInfo env)
```
rightfold: When starting up, we should log whether authorization is enabled, and whether a secret has been given (but not the actual key). This because not enabling authorization is a possible security problem.

server/src/JwtMiddleware.hs
```diff
@@ -0,0 +1,116 @@
...
+                  -> AuthRejected OperationNotAllowed
+
+-- | Extract the JWT claim from the request.
+getRequestClaim :: Http.RequestHeaders -> Http.Query -> POSIXTime -> Maybe JWT.Secret -> Either TokenError IcepeakClaim
+getRequestClaim headers query now maybeSecret =
+  let getTokenBytes = maybe (Left $ VerificationError TokenNotFound) Right (findTokenBytes headers query)
+  in case maybeSecret of
+       -- authorization is enabled, but no secret provided, accept all tokens
+       Nothing     -> getTokenBytes >>= extractClaimUnverified
+       Just secret -> getTokenBytes >>= extractClaim now secret
```
rightfold: Make more clear that the comment only applies to the `Nothing` case. For example by moving it in there, or by moving the `Just` case to above the comment.

server/integration-tests/auth_stress_test.hs
```diff
@@ -0,0 +1,48 @@
...
+
+main :: IO ()
+main = do
+  args <- getArgs
+  case args of
+    [count] -> either fail main' (readEither count)
+    _ -> fail "usage: stress_test.hs <count>"
+  where fail msg = do { hPutStrLn stderr msg; exitFailure }
+
+testToken :: SBS.ByteString
```
NunoAlexandre: I personally am I a fan of placing code after the place where it is called, so to get a top-down view of the code, i.e, on top is the overview of what's going on and the details follow down below. Thus `testToken` should go under `main'`.

server/integration-tests/generate_test_token.hs
```diff
@@ -0,0 +1,40 @@
...
+
+import qualified Data.ByteString as SBS
+
+import AccessControl
+import JwtAuth
+import qualified Data.Map.Strict as Map
+import qualified Web.JWT as JWT
+import qualified Data.Text.IO as Text
+
+testAccess :: IcepeakClaim
```
NunoAlexandre: Same as above: `testAccess` should go under `testClaims`.
And `main` should be on top.

server/package.yaml
```diff
@@ -16,6 +16,7 @@ dependencies:
 - directory >= 1.3 && <1.4
 - hashable >=1.2 && <1.3
 - http-types >=0.9 && <0.10
+- jwt >= 0.7 && <0.8
```
NunoAlexandre: Doesn't `<0.8` pick the latest version possible smaller than 0.8? Is `>=0.7` redundant?

server/package.yaml
```diff
@@ -43,10 +45,12 @@ library:
   - -fno-ignore-asserts
   - -funbox-strict-fields
   exposed-modules:
+  - AccessControl
   - Config
   - Core
-  - Guardian
```
NunoAlexandre: Goodbye Guardian :heart:

server/src/AccessControl.hs
```diff
@@ -0,0 +1,97 @@
...
+import           Data.Text  (Text)
+
+import           Store      (Path)
+
+-- * Claim datatypes
+
+-- | Defines the structure of a JWT claim for Icepeak.
+data IcepeakClaim = IcepeakClaim
+  { icepeakClaimWhitelist :: [AuthPath]
+    -- ^ The whitelist containing all authorizations.
```
NunoAlexandre: if you need this comment the name is probably not good enough. Would rename it to`icepeakClaimAcessWhitelist` make this comment die?

server/src/AccessControl.hs
```diff
@@ -0,0 +1,97 @@
...
+import           Data.Aeson ((.:), (.=))
+import qualified Data.Aeson as Aeson
+import qualified Data.List  as List
+import           Data.Text  (Text)
+
+import           Store      (Path)
+
+-- * Claim datatypes
+
+-- | Defines the structure of a JWT claim for Icepeak.
```
NunoAlexandre: This comment is also just noise, IMO.

server/src/AccessControl.hs
```diff
@@ -0,0 +1,97 @@
...
+-- | A claim that allows all operations.
+allowEverything :: IcepeakClaim
+allowEverything = IcepeakClaim [AuthPath [] [minBound..maxBound]]
+
+-- * Authorization
+
+-- | Check whether accessing the given path with the given mode is authorized by
+-- the supplied claim.
+isAuthorizedByClaim :: IcepeakClaim -> Path -> AccessMode -> Bool
+isAuthorizedByClaim claim path mode = any allows (icepeakClaimWhitelist claim) where
```
NunoAlexandre: I'd say to move `where` to the next line to be consistent with the rest of the project.

server/src/AccessControl.hs
```diff
@@ -0,0 +1,97 @@
...
+  allows (AuthPath prefix modes) = List.isPrefixOf prefix path && mode `elem` modes
+
+
+-- * JSON encoding and decoding
+
+accessModeToText :: AccessMode -> Text
+accessModeToText mode = case mode of
+    ModeRead   -> "read"
```
NunoAlexandre: Wouldn't it be better to `instance` `tshow`? I think @rightfold has done it somewhere else.

NunoAlexandre: Also, note the incnsistent spacing:
``` haskell
ModeRead   -> "read"
ModeWrite -> "write"
```

I don't really care between right aligning or not, but it's good having it consistent - at least within a function.

server/src/Config.hs
```diff
@@ -3,28 +3,49 @@ module Config (
...
+    -- specified even though JWT authorization is enabled, tokens will still be
+    -- used, but not be verified.
+    configJwtSecret :: Maybe JWT.Secret
 }
 
 -- Parsing of command-line arguments
 
-configParser :: Parser Config
-configParser =
+configParser :: [(String, String)] -> Parser Config
```
NunoAlexandre: Maybe add a type alias out of `[(String, String)]` or wrap it in a `newtype`?

server/src/Guardian.hs
```diff
@@ -1,27 +0,0 @@
-{-# LANGUAGE OverloadedStrings #-}
```
NunoAlexandre: giphy

server/src/HttpServer.hs
```diff
@@ -2,38 +2,39 @@
...
+import Core (Core (..), EnqueueResult (..))
+import Config (Config (..))
 import qualified Core
 
-
 new :: Core -> IO Application
 new core =
   scottyApp $ do
+    when (configEnableJwtAuth $ coreConfig core) $
+      middleware $ jwtMiddleware $ configJwtSecret $ coreConfig core
```
NunoAlexandre: So many dollars :speak_no_evil: 
It doesn't really matter but I guess most of us would do:
``` haskell
(middleware . jwtMiddleware . configJwtSecret . coreConfig) core
```

server/src/JwtAuth.hs
```diff
@@ -0,0 +1,103 @@
...
+   Nothing     -> Left TokenSignatureInvalid
+   Just token' -> Right token'
+
+decodeToken :: SBS.ByteString -> Either VerificationError (JWT UnverifiedJWT)
+decodeToken bytes =
+ case JWT.decode (Text.decodeUtf8 bytes) of
+   Nothing    -> Left TokenInvalid
+   Just token -> Right token
+
+-- | Check that a token is valid at the given time for the given secret.
```
NunoAlexandre: This should be at the top of the module.

server/tests/JwtSpec.hs
```diff
@@ -0,0 +1,68 @@
...
+          }
+
+    let testClaims = addIcepeakClaim testAccess emptyClaim
+
+    let testSecret = JWT.binarySecret "2o8357cEuc2o835cmsoei"
+
+    let now = 12512 :: NominalDiffTime
+
+    it "should accept a valid token" $ do
+      True
```
NunoAlexandre: Cheating!

server/app/Icepeak.hs
fatho: Done!

server/src/JwtMiddleware.hs
fatho: Fixed!

PR #13, opened 2017-09-26
Add Prometheus metrics endpoint

### LINKED ISSUE #10: Metrics
Icepeak should publish metrics so that [Prometheus][p] can read them.

It should do this by listening for HTTP requests on a port separate from the port used by user requests. Prometheus will send requests to retrieve the metrics whenever it wants to; Prometheus is pull-based.

This feature should be optional: it may be enabled only if starting with a `--metrics <host>:<port>` option or similar environment variable.

It should publish at least the following:

 - Total size of the data stored, so that one can be alerted when it increases beyond normal size. We can just stat the data file for this (near constant time); no need to traverse the JSON data (linear time). This should be a Prometheus "Gauge".
 - Total number of PUTs and DELETEs since starting Icepeak. Prometheus can itself split this into units of time. This should be a Prometheus "Counter".
 - Might as well add GHC metrics, it's one line of code.

You can use the following packages:

 - [prometheus-client][pclient] for keeping the metrics in memory.
 - [prometheus-metrics-ghc][pghc] for gathering the GHC metrics.
 - [wai-middleware-prometheus][pwaimw] (`metricsApp` application) for publishing the metrics.

[p]: https://prometheus.io/
[pghc]: https://hackage.haskell.org/package/prometheus-metrics-ghc
[pclient]: https://hackage.haskell.org/package/prometheus-client
[pwaimw]: https://hackage.haskell.org/package/wai-middleware-prometheus

fatho:
Fixes #10.

### DIFF
--- a/server/app/Icepeak.hs
+++ b/server/app/Icepeak.hs
@@ -4,15 +4,19 @@ module Main where
 
 import Control.Concurrent.Async
 import Control.Exception (try, SomeException)
-import Control.Monad (void)
+import Control.Monad (forM, void)
 import Data.Aeson (eitherDecodeStrict, Value)
 import Data.ByteString (hGetContents, ByteString)
+import Data.Foldable (forM_)
 import Options.Applicative (execParser)
 import Prelude hiding (log)
-import System.IO (withFile, IOMode (..))
 import System.Environment (getEnvironment)
+import System.IO (withFile, IOMode (..))
 
 import qualified Control.Concurrent.Async as Async
+import qualified Data.ByteString as SBS
+import qualified Prometheus
+import qualified Prometheus.Metric.GHC
 import qualified System.Posix.Signals as Signals
 
 import Config (Config (..), configInfo)
@@ -23,6 +27,8 @@ import qualified Core
 import qualified HttpServer
 import qualified Server
 import qualified WebsocketServer
+import qualified Metrics
+import qualified MetricsServer
 
 -- Install SIGTERM and SIGINT handlers to do a graceful exit.
 installHandlers :: Core -> Async () -> IO ()
@@ -40,30 +46,41 @@ installHandlers core serverThread =
     void $ installHandler Signals.sigINT
 
 
-readValue :: FilePath -> IO Value
-readValue filePath = do
+readValue :: FilePath -> Maybe Metrics.IcepeakMetrics -> IO Value
+readValue filePath metrics = do
   eitherEncodedValue <- try $ withFile filePath ReadMode hGetContents
   case (eitherEncodedValue :: Either SomeException ByteString) of
       Left exc -> error $ "Failed to read the data from disk: " ++ show exc
-      Right encodedValue -> case eitherDecodeStrict encodedValue of
-        Left msg  -> error $ "Failed to decode the initial data: " ++ show msg
-        Right value -> return value
-
+      Right encodedValue -> do
+        forM_ metrics $ Metrics.setDataSize (SBS.length encodedValue)
+        case eitherDecodeStrict encodedValue of
+          Left msg  -> error $ "Failed to decode the initial data: " ++ show msg
+          Right value -> return value
 
 main :: IO ()
 main = do
   env <- getEnvironment
   config <- execParser (configInfo env)
+
+  -- setup metrics if enabled
+  icepeakMetrics <- forM (configMetricsEndpoint config) $ const $ do
+    void $ Prometheus.register Prometheus.Metric.GHC.ghcMetrics
+    Metrics.createAndRegisterIcepeakMetrics
+
   -- load the persistent data from disk
   let filePath = configDataFile config
-  value <- readValue filePath
-  core <- Core.newCore value config
+  value <- readValue filePath icepeakMetrics
+  core <- Core.newCore value config icepeakMetrics
   httpServer <- HttpServer.new core
   let wsServer = WebsocketServer.acceptConnection core
+
+  -- start threads
   pops <- Async.async $ Core.processOps core
   upds <- Async.async $ WebsocketServer.processUpdates core
   serv <- Async.async $ Server.runServer wsServer httpServer
   logger <- Async.async $ processLogRecords (coreLogRecords core)
+  metrics <- Async.async
+    $ forM_ (configMetricsEndpoint config) MetricsServer.runMetricsServer
   installHandlers core serv
   logAuthSettings config (coreLogRecords core)
   log "System online. ** robot sounds **" (coreLogRecords core)
@@ -73,6 +90,8 @@ main = do
   void $ Async.wait upds
   void $ Async.wait serv
   void $ Async.wait logger
+  -- kill metrics endpoint when the main threads ended
+  Async.cancel metrics
 
 logAuthSettings :: Config -> LogQueue -> IO ()
 logAuthSettings cfg queue
--- a/server/src/Config.hs
+++ b/server/src/Config.hs
@@ -1,46 +1,64 @@
 module Config (
   Config (..),
-  configInfo
+  MetricsConfig (..),
+  configInfo,
 ) where
 
 import Control.Applicative (optional)
-import Data.Semigroup      ((<>))
-import Data.List           (lookup)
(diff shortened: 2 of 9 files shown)


server/src/Config.hs
```diff
@@ -1,46 +1,72 @@
...
+metricsServerConfig cfg = Warp.defaultSettings
+  & Warp.setHost (metricsConfigHost cfg)
+  & Warp.setPort (metricsConfigPort cfg)
+
+metricsConfigReader :: ReadM MetricsConfig
+metricsConfigReader = eitherReader $ \input ->
+  case List.findIndex (== ':') input of
+    Nothing -> Left "no port specified"
+    Just idx -> let (hostStr, portStr) = List.splitAt idx input
+                in MetricsConfig (fromString hostStr) <$> Read.readEither (drop 1 portStr)
```
NunoAlexandre: Could also be:

``` haskell
metricsConfigReader = eitherReader $ \input ->
  let (hostStr, portStr) = Data.List.break (== ':') input in  
  case List.findIndex (== ':') input of
     (_, "") -> Left "no port specified"
     Just (hostStr, portStr) -> MetricsConfig (fromString hostStr) <$> Read.readEither (drop 1 portStr)
```

server/src/Core.hs
```diff
@@ -124,6 +128,9 @@ processOps core = go Null
           -- temporary file will thus not corrupt the datastore
           writeFile tempFileName (encodeToLazyText newValue)
           renameFile tempFileName fileName
+          forM_ (coreMetrics core) $ \m -> do
+            stat <- Posix.getFileStatus fileName
+            Metrics.setDataSize (Posix.fileSize stat) m
```
NunoAlexandre: Maybe extract this into a separate function?

server/src/HttpServer.hs
```diff
@@ -21,17 +23,20 @@ new core =
       middleware $ jwtMiddleware $ configJwtSecret $ coreConfig core
 
     get (regex "^") $ do
+      liftIO $ forM_ (coreMetrics core) Metrics.notifyGetRequest
```
NunoAlexandre: Can you abstract this to something like:

``` haskell
updateMetrics :: Metrics.Notification -> ?
updateMetrics = liftIO . forM_ (coreMetrics core) 

-- and then:
put (regex "^") $ do
       updateMetrics Metrics.notifyPutRequest
--etc

The main point is that we want it to read high-level, like prose. 
```

server/src/MetricsServer.hs
```diff
@@ -0,0 +1,8 @@
+module MetricsServer where
+
+import qualified Network.Wai.Handler.Warp          as Warp
+import qualified Network.Wai.Middleware.Prometheus as PrometheusWai
+
+runMetricsServer :: Warp.Settings -> IO ()
+runMetricsServer settings = do
+  Warp.runSettings settings PrometheusWai.metricsApp
```
NunoAlexandre: I am ok with having a small module like this, but maybe more stuff could be moved here?

server/src/WebsocketServer.hs
```diff
@@ -93,9 +94,11 @@ handleClient conn path core = do
     logRecords = coreLogRecords core
     onConnect = do
       modifyMVar_ state (pure . Subscription.subscribe path uuid conn)
+      forM_ (coreMetrics core) Metrics.incrementSubscribers
```
NunoAlexandre: Make it read as prose.

server/src/WebsocketServer.hs
```diff
@@ -93,9 +94,11 @@ handleClient conn path core = do
...
       log (T.pack $ "Client " ++ (show uuid) ++ " connected, subscribed to " ++ (show path) ++ ".") logRecords
     onDisconnect = do
       modifyMVar_ state (pure . Subscription.unsubscribe path uuid)
+      forM_ (coreMetrics core) Metrics.decrementSubscribers
```
NunoAlexandre: Make it read as prose.

server/app/Icepeak.hs
```diff
@@ -40,30 +47,43 @@ installHandlers core serverThread =
...
+        forM_ metrics $ Metrics.setDataSize (SBS.length encodedValue)
+        case eitherDecodeStrict encodedValue of
+          Left msg  -> error $ "Failed to decode the initial data: " ++ show msg
+          Right value -> return value
 
 main :: IO ()
 main = do
   env <- getEnvironment
   config <- execParser (configInfo env)
+
```
NunoAlexandre: This `main` function is growing places. I would be OK with it being long if the levels of abstraction it included were consistent, i.e, if it was just glueing things together, but there are some lower level stuff here that should be packed into a module IMO.

NunoAlexandre:
@fatho another question:

In WM the metrics are NOT invasive, in the sense we didn't need to change any function signature and manually build the metrics. Why is it necessary here? Is it because we need more stuff than what we get out of the box?

fatho:
@NunoAlexandre They are invasive because we not only have GHC metrics, but also want to measure metrics specific to the application (i.e. file size, number of requests, etc.), and those need to be recorded somewhere.

server/src/Config.hs
```diff
@@ -1,46 +1,72 @@
...
+  , configMetricsEndpoint :: Maybe MetricsConfig
+  }
+
+data MetricsConfig = MetricsConfig
+  { metricsConfigHost :: Warp.HostPreference
+  , metricsConfigPort :: Warp.Port
+  }
+
+metricsServerConfig :: MetricsConfig -> Warp.Settings
+metricsServerConfig cfg = Warp.defaultSettings
```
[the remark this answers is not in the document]
fatho: The Warp module does not export the constructor or record accessors of the `Settings` type. Maybe in some internal module, but using the provided getters and setters is (more) future-proof.

server/src/Config.hs
fatho: Using `break` is a good idea, although there's something wrong with your code snippet.

server/src/Core.hs
fatho: Since this is only used once, making it a top level function seems to introduce quite a lot of overhead (because all the local variables would need to be passed as arguments). I'd rather leave it as it is.

server/src/HttpServer.hs
fatho: Done, does indeed look nicer now.

server/src/MetricsServer.hs
fatho: Initially, I thought there would be more to the server, but it is indeed that simple. I think I'll move the function that builds the actual warp settings there.

server/src/WebsocketServer.hs
fatho: Done

server/src/WebsocketServer.hs
fatho: Done

server/app/Icepeak.hs
fatho: I moved the log output for metrics to another place. I think overall, the main function is just gluing stuff together. First, it creates the stateful "objects", then it starts the thread, then it waits.

Loading the data file might be placed into the Core instead, but I think that is it out of the scope of this PR.

server/src/Core.hs
NunoAlexandre: Ok. The argument "this is only used once" is no argument since most functions are only used once and that would be no excuse to have the whole program _in_ `main` ;) I see what you get however, altough small reasonable points like that make programs grow back hairs.

server/src/HttpServer.hs
NunoAlexandre: Nice :+1:. I find this to be a cornerstone of my programming culture: to make code read as prose.

server/app/Icepeak.hs
NunoAlexandre: Yeah, it is out of scope. But still, note that if `main` knows how to build some stateful "objects", it is not just glueing stuff.

NunoAlexandre:
> @NunoAlexandre They are invasive because we not only have GHC metrics, but also want to measure metrics specific to the application (i.e. file size, number of requests, etc.), and those need to be recorded somewhere.

Ok, that's it, it's clear. Thanks!

PR #14, opened 2017-09-28
Rename metrics and provide more detailed HTTP request metrics

fatho:


fatho:
Since this is a breaking change, we should probably also bump the version to `0.2.0`, right?

server/src/Metrics.hs
```diff
@@ -1,31 +1,27 @@
 module Metrics where
 
 import           Prometheus
+import qualified Network.HTTP.Types as Http
+import qualified Data.ByteString.Char8 as BS8
+
+type RequestCounter = Vector (String, String) Counter
```
NunoAlexandre: Looking at this for the first time, I have no idea what this is actually counting. What will the values of the pairs be? 

Try to make the code so that it answers these questions itself alone. If not possible at all, leave a comment.

server/src/Metrics.hs
```diff
@@ -1,31 +1,27 @@
...
-notifyGetRequest :: IcepeakMetrics -> IO ()
-notifyGetRequest = incCounter . icepeakMetricsGetCounter
-
-notifyPutRequest :: IcepeakMetrics -> IO ()
-notifyPutRequest = incCounter . icepeakMetricsPutCounter
-
-notifyDeleteRequest :: IcepeakMetrics -> IO ()
-notifyDeleteRequest = incCounter . icepeakMetricsDeleteCounter
+  <$> registerIO (vector ("method", "status") $ counter (Info "icepeak_http_requests" "Total number of HTTP requests since starting Icepeak."))
+  <*> registerIO (gauge (Info "icepeak_data_size" "Size of data file in bytes."))
```
NunoAlexandre: Can you make it shorter?

``` haskell
createAndRegisterIcepeakMetrics = IcepeakMetrics
  <$> registerIO (vector ("method", "status") 
        $ counter (Info "icepeak_http_requests" "Total number of HTTP requests since starting Icepeak."))
  <*> registerIO (gauge (Info "icepeak_data_size" "Size of data file in bytes."))
  <*> registerIO (gauge (Info "icepeak_subscriber_count" "Number of websocket subscriber connections."))
```

or 

``` haskell
createAndRegisterIcepeakMetrics = IcepeakMetrics
  <$> registerIO (vector ("method", "status")  requestCounter
  <*> registerIO (gauge (Info "icepeak_data_size" "Size of data file in bytes."))
  <*> registerIO (gauge (Info "icepeak_subscriber_count" "Number of websocket subscriber connections."))
    where requestCounter = counter (Info "icepeak_http_requests" "Total number of HTTP requests since starting Icepeak."))
```

server/src/Metrics.hs
```diff
@@ -1,31 +1,27 @@
...
+  <*> registerIO (gauge (Info "icepeak_subscriber_count" "Number of websocket subscriber connections."))
+
+notifyRequest :: Http.Method -> Http.Status -> IcepeakMetrics -> IO ()
+notifyRequest method status =
+  let lbl = (BS8.unpack method, show $ Http.statusCode status)
```
NunoAlexandre: 1. I don't mind if we use `let` or `where` (I prefer `where` though) but we should try to be more consistent.

2. `s/lbl/label`. _say no to abbreviations_ : https://markus.oberlehner.net/blog/abbreviations-suck/

server/src/Metrics.hs
```diff
@@ -14,14 +17,17 @@ data IcepeakMetrics = IcepeakMetrics
...
+  where
+    requestCounter = counter (Info "icepeak_http_requests"
+                                   "Total number of HTTP requests since starting Icepeak.")
 
 notifyRequest :: Http.Method -> Http.Status -> IcepeakMetrics -> IO ()
-notifyRequest method status =
-  let lbl = (BS8.unpack method, show $ Http.statusCode status)
-  in withLabel lbl incCounter . icepeakMetricsRequestCounter
+notifyRequest method status = withLabel label incCounter . icepeakMetricsRequestCounter
+  where
```
NunoAlexandre: I would inline this, but doesn't really matter.

server/src/Metrics.hs
```diff
@@ -4,7 +4,10 @@ import           Prometheus
 import qualified Network.HTTP.Types as Http
 import qualified Data.ByteString.Char8 as BS8
 
-type RequestCounter = Vector (String, String) Counter
+type MethodLabel = String
+type StatusLabel = String
+
+type RequestCounter = Vector (MethodLabel, StatusLabel) Counter
```
NunoAlexandre: Looks better @fatho! Again, as an outsider, I wonder if `Method` is `HttpMethod` or access method (read / write)?

server/src/Metrics.hs
```diff
@@ -4,13 +4,16 @@ import           Prometheus
...
-type MethodLabel = String
-type StatusLabel = String
+type HttpMethodLabel = String
+type HttpStatusLabel = String
 
-type RequestCounter = Vector (MethodLabel, StatusLabel) Counter
+type HttpRequestCounter = Vector (HttpMethodLabel, HttpStatusLabel) Counter
+
+countHttpRequest :: Http.Method -> Http.Status -> Metric HttpRequestCounter -> IO ()
+countHttpRequest method status = withLabel (BS8.unpack method, show $ Http.statusCode status) incCounter
```
NunoAlexandre: Sorry man, I meant to inline the `where` instead of having it in a new line:

``` haskell
foo = bar
  where bar = this
```

Instead of

``` haskell
foo = bar
  where
     bar = this
```

server/src/Metrics.hs
```diff
@@ -4,13 +4,16 @@ import           Prometheus
 import qualified Network.HTTP.Types as Http
 import qualified Data.ByteString.Char8 as BS8
 
-type MethodLabel = String
-type StatusLabel = String
+type HttpMethodLabel = String
+type HttpStatusLabel = String
 
-type RequestCounter = Vector (MethodLabel, StatusLabel) Counter
+type HttpRequestCounter = Vector (HttpMethodLabel, HttpStatusLabel) Counter
```
NunoAlexandre: Loud and clear! :+1:

server/src/Metrics.hs
NunoAlexandre: Ok, I see that you created this from the where clause of the other function. I think it's alright, but move this function underneath the `notifyRequest` function so improve code proximity and reading travel effort.

fatho: In terms of code proximity, I'd keep it this way, `countHttpRequest` is more general and works on `HttpRequestCounter` metrics, while all the other functions below are working on the aggregate `IcepeakMetrics`.

NunoAlexandre: Aright, fair enough.

PR #15, opened 2017-10-02
Add metric for the total amount of bytes written to disk

fatho:


rightfold:
And don't forget to tag.

PR #16, opened 2017-10-04
Make queue capacity configurable

rightfold:


server/src/Config.hs
```diff
@@ -59,6 +61,11 @@ configParser environment = Config
                   metavar "HOST:PORT" <>
                   help "If provided, Icepeak collects various metrics and provides them to Prometheus on the given endpoint."
                  ))
+  <*> fmap (fromMaybe 256)
```
fatho: It's possible to add a default value with the `value` modifier, i.e.

```
(option auto
                       (long "queue-capacity" <>
                        metavar "INTEGER" <>
                        value 256 <>
                        help "Higher values allow more requests to come in within a short amount of time.")
```

looks a bit cleaner then.

rightfold: Ah, nice. :+1:

server/src/Config.hs
```diff
@@ -59,6 +61,11 @@ configParser environment = Config
...
+           (optional (option auto
+                       (long "queue-capacity" <>
+                        metavar "INTEGER" <>
+                        help "Higher values allow more requests to come in within a short amount of time.")))
```
ruuda: Maybe mention that this is a trade-off between the amount of data risked losing, the ability to accept bursts of requests without denying them.

rightfold: Good point. :+1:

server/src/Config.hs
```diff
@@ -59,6 +60,12 @@ configParser environment = Config
                   metavar "HOST:PORT" <>
                   help "If provided, Icepeak collects various metrics and provides them to Prometheus on the given endpoint."
                  ))
+  <*> option auto
+       (long "queue-capacity" <>
+        metavar "INTEGER" <>
+        value 256 <>
+        help ("Higher values allow more requests to come in within a short amount of time. " <>
+              "Smaller values increase the risk of data loss."))
```
rkrzr: s/increase/decrease

If the queue is smaller, fewer values can be lost.

rightfold: Right, but you also lose more requests. Will clarify and merge.

PR #17, opened 2017-10-04
Fix that all data would be lost when restarting

fatho:


server/integration-tests/data-survives-restart.sh
```diff
@@ -0,0 +1,47 @@
...
+
+if [ ! -f "stack.yaml" ]; then
+    echo "Run test in server package directory"
+    exit 1
+fi
+
+# PERARATIONS
+stack build
+DATA_FILE=`mktemp`
+echo '{}' > $DATA_FILE
```
rightfold: Might want to quote this, since it is also quoted elsewhere:

```bash
echo '{}' > "$DATA_FILE"
```

server/integration-tests/data-survives-restart.sh
```diff
@@ -0,0 +1,47 @@
+#/bin/bash
+
+if [ ! -f "stack.yaml" ]; then
+    echo "Run test in server package directory"
+    exit 1
+fi
+
+# PERARATIONS
```
rkrzr: s/PERARATIONS/Preparations

server/integration-tests/data-survives-restart.sh
fatho: Done

server/integration-tests/data-survives-restart.sh
fatho: Done

Issue #19, opened 2017-10-05
Do not write the entire data set on each request


rightfold:
This is obviously inefficient and bad. We should instead write periodically, by having a queue of commands instead of operations, where a command is either an operation or "sync". A timer posts "sync" commands every N seconds.

If we also want to guarantee durability (we currently don't), then we want requests to block until the operation has succeeded, and we should append operations to a journal (we don't want requests to wait for the periodic sync). The sync command would then clear the journal when writing the data to disk. Restarting Icepeak will restore from the journal.

fatho:
Both syncing and journaling have been implemented. A new issue has been filed for blocking requests #25.

PR #20, opened 2017-10-05
Implement periodic syncing of database to disk

fatho:
Addresses first part of #19.

### DIFF
--- a/server/app/Icepeak.hs
+++ b/server/app/Icepeak.hs
@@ -3,19 +3,14 @@
 module Main where
 
 import Control.Concurrent.Async
-import Control.Exception (try, SomeException)
 import Control.Monad (forM, void)
-import Data.Aeson (eitherDecodeStrict, Value)
-import Data.ByteString (hGetContents, ByteString)
 import Data.Foldable (forM_)
 import Data.Semigroup ((<>))
 import Options.Applicative (execParser)
 import Prelude hiding (log)
 import System.Environment (getEnvironment)
-import System.IO (withFile, IOMode (..))
 
 import qualified Control.Concurrent.Async as Async
-import qualified Data.ByteString as SBS
 import qualified Data.Text as Text
 import qualified Prometheus
 import qualified Prometheus.Metric.GHC
@@ -37,9 +32,9 @@ installHandlers :: Core -> Async () -> IO ()
 installHandlers core serverThread =
   let
     handle = do
+      log "\nTermination sequence initiated ..." (coreLogRecords core)
       Core.postQuit core
       Async.cancel serverThread
-      log "\nTermination sequence initiated ..." (coreLogRecords core)
     handler = Signals.CatchOnce handle
     blockSignals = Nothing
     installHandler signal = Signals.installHandler signal handler blockSignals
@@ -47,17 +42,6 @@ installHandlers core serverThread =
     void $ installHandler Signals.sigTERM
     void $ installHandler Signals.sigINT
 
-
-readValue :: FilePath -> Maybe Metrics.IcepeakMetrics -> IO Value
-readValue filePath metrics = do
-  eitherEncodedValue <- try $ withFile filePath ReadMode hGetContents
-  case (eitherEncodedValue :: Either SomeException ByteString) of
-      Left exc -> error $ "Failed to read the data from disk: " ++ show exc
-      Right encodedValue -> do
-        forM_ metrics $ Metrics.setDataSize (SBS.length encodedValue)
-        case eitherDecodeStrict encodedValue of
-          Left msg  -> error $ "Failed to decode the initial data: " ++ show msg
-          Right value -> return value
 
 main :: IO ()
 main = do
@@ -69,15 +53,18 @@ main = do
     void $ Prometheus.register Prometheus.Metric.GHC.ghcMetrics
     Metrics.createAndRegisterIcepeakMetrics
 
-  -- load the persistent data from disk
-  let filePath = configDataFile config
-  value <- readValue filePath icepeakMetrics
-  core <- Core.newCore value config icepeakMetrics
+  eitherCore <- Core.newCore config icepeakMetrics
+  either putStrLn runCore eitherCore
+
+runCore :: Core -> IO ()
+runCore core = do
+  let config = coreConfig core
   httpServer <- HttpServer.new core
   let wsServer = WebsocketServer.acceptConnection core
 
   -- start threads
-  pops <- Async.async $ Core.processOps core
+  pops <- Async.async $ Core.runCommandLoop core
+  sync <- Async.async $ Core.runSyncTimer core
   upds <- Async.async $ WebsocketServer.processUpdates core
   serv <- Async.async $ Server.runServer wsServer httpServer
   logger <- Async.async $ processLogRecords (coreLogRecords core)
@@ -86,6 +73,7 @@ main = do
   installHandlers core serv
   logAuthSettings config (coreLogRecords core)
   logQueueSettings config (coreLogRecords core)
+  logSyncSettings config (coreLogRecords core)
   log "System online. ** robot sounds **" (coreLogRecords core)
 
   -- TODO: Log exceptions properly (i.e. non-interleaved)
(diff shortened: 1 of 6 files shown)


server/src/Config.hs
```diff
@@ -66,6 +62,13 @@ configParser environment = Config
         value 256 <>
         help ("Smaller values decrease the risk of data loss during a crash, while " <>
               "higher values result in more requests being accepted in rapid succession."))
+  <*> optional (option timeDurationReader
+       (long "sync-interval" <>
+        metavar "DURATION" <>
+        help ("If supplied, data is only persisted to disc every DURATION time units." <>
+              "The units 'm' (minutes), 's' (seconds), 'ms' (milliseconds) " <>
+              "and 'mus' (microseconds) can be used. " <>
```
rightfold: I don't think microsecond intervals are useful.

rkrzr: Yes, I would drop the microseconds here.

fatho: Probably not, I can remove it if desired. But it doesn't hurt I think, because one always has to specify a unit. It cannot happen that someone accidentally syncs every 5 microseconds.

server/src/Core.hs
```diff
@@ -74,70 +90,116 @@ type ServerState = SubscriptionTree UUID WS.Connection
...
+    writeTBQueue (coreQueue core) Stop
     writeTBQueue (coreUpdates core) Nothing
     writeTBQueue (coreLogRecords core) Nothing
 
-enqueueOp :: Op -> Core -> IO EnqueueResult
-enqueueOp op core = atomically $ do
+-- | Try to enqueue a command. It succeeds if the queue is not full, otherwise,
+-- nothing is changed.
+tryEnqueueCommand :: Command -> Core -> IO EnqueueResult
+tryEnqueueCommand op core = atomically $ do
```
NunoAlexandre: can you now rename `op` to `cmd` or `command`?

server/src/Core.hs
```diff
@@ -74,70 +90,116 @@ type ServerState = SubscriptionTree UUID WS.Connection
...
   isFull <- isFullTBQueue (coreQueue core)
-  unless isFull $ writeTBQueue (coreQueue core) (Just op)
+  unless isFull $ writeTBQueue (coreQueue core) op
   pure $ if isFull then Dropped else Enqueued
 
+-- | Enqueue a command. Blocks if the queue is full.
+enqueueCommand :: Command -> Core -> IO ()
+enqueueCommand cmd core = atomically $ writeTBQueue (coreQueue core) cmd
```
NunoAlexandre: When/why should this be used instead of `tryEnqueueCommand`?

server/src/Core.hs
```diff
@@ -74,70 +90,116 @@ type ServerState = SubscriptionTree UUID WS.Connection
...
 
 withCoreMetrics :: MonadIO m => Core -> (Metrics.IcepeakMetrics -> IO ()) -> m ()
 withCoreMetrics core act = liftIO $ forM_ (coreMetrics core) act
 
--- Execute an operation.
-handleOp :: Op -> Value -> Value
-handleOp op value = case op of
+-- Execute a modification.
+applyModification :: Modification -> Value -> Value
+applyModification op value = case op of
```
NunoAlexandre: s/op/modification

server/src/Core.hs
```diff
@@ -74,70 +90,116 @@ type ServerState = SubscriptionTree UUID WS.Connection
...
+-- queue.
+runCommandLoop :: Core -> IO Value
+runCommandLoop core = atomically (readTVar (coreCurrentValue core)) >>= go
   where
     go val = do
-      maybeOp <- atomically $ readTBQueue (coreQueue core)
-      case maybeOp of
-        Just op -> do
-          let newValue = handleOp op val
+      command <- atomically $ readTBQueue (coreQueue core)
```
NunoAlexandre: `atomically $ readTBQueue (coreQueue core)` shouldn't be _naked_ here. Is there no function already doing this? It seems like it should.

server/src/Core.hs
```diff
@@ -74,70 +90,116 @@ type ServerState = SubscriptionTree UUID WS.Connection
...
-          -- we first write to a temporary file here and then do a rename on it
-          -- because rename is atomic on Posix and a crash during writing the
-          -- temporary file will thus not corrupt the datastore
-          writeFile tempFileName (encode newValue)
-          renameFile tempFileName fileName
-          forM_ (coreMetrics core) $ \m -> do
-            stat <- Posix.getFileStatus fileName
-            Metrics.setDataSize (Posix.fileSize stat) m
-            Metrics.incrementDataWritten (Posix.fileSize stat) m
+            writeTVar (coreValueIsDirty core) True
```
NunoAlexandre: Minor, but I'd prefer to say `setCoreValueIsDirty` instead of `writeTVar (coreValueIsDirty core) True`.

server/src/Core.hs
```diff
@@ -74,70 +90,116 @@ type ServerState = SubscriptionTree UUID WS.Connection
...
 withCoreMetrics :: MonadIO m => Core -> (Metrics.IcepeakMetrics -> IO ()) -> m ()
 withCoreMetrics core act = liftIO $ forM_ (coreMetrics core) act
 
--- Execute an operation.
-handleOp :: Op -> Value -> Value
-handleOp op value = case op of
+-- Execute a modification.
+applyModification :: Modification -> Value -> Value
+applyModification op value = case op of
   Delete path -> Store.delete path value
```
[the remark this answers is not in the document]
fatho: Good suggestion!

server/src/Core.hs
fatho: Done

server/src/Core.hs
fatho: Added explanation

server/src/Core.hs
fatho: Done

server/src/Core.hs
```diff
@@ -61,7 +76,8 @@ data EnqueueResult = Enqueued | Dropped
 
 data Core = Core
   { coreCurrentValue :: TVar Value
-  , coreQueue :: TBQueue (Maybe Op)
+  , coreValueIsDirty :: TVar Bool
```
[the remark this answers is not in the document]
fatho: Added explanation

server/src/Core.hs
fatho: I like it this way. It makes it immediately visible that this is an STM transaction (that could possibly be extended with more actions).

server/src/Core.hs
fatho: Same here, this conveys more information in my opinion.

server/src/Core.hs
```diff
@@ -3,53 +3,67 @@ module Core
...
 -- A modification operation.
-data Op
+data Modification
   = Put Path Value
   | Delete Path
   deriving (Eq, Show)
 
-opPath :: Op -> Path
-opPath op = case op of
+data Command
```
rkrzr: Please document these commands.

server/app/Icepeak.hs
```diff
@@ -86,15 +73,17 @@ main = do
...
 
   -- TODO: Log exceptions properly (i.e. non-interleaved)
   void $ Async.wait pops
   void $ Async.wait upds
   void $ Async.wait serv
   void $ Async.wait logger
-  -- kill metrics endpoint when the main threads ended
+  -- kill auxiliary threads when the main threads ended
   Async.cancel metrics
+  Async.cancel sync
```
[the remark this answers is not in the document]
fatho:
```
  void $ Async.wait pops
```
This is the core loop that is stopped when `Ctrl+C`ing by posting the `Stop` command https://github.com/channable/icepeak/blob/9b296d87b4f4afad0208594409296576966ca8a2/server/app/Icepeak.hs#L36. Therefore, in main we simply wait until it's time to stop.

```
  void $ Async.wait upds
```
This is the websocket loop, which receives it's cancellation via the subscription queue from the queue, therefore we also just need to wait until it's done.

```
  void $ Async.wait serv
```
The `Ctrl+C` handler also `Async.cancel`s the server thread (because any requests coming in afterwards would have no effect, so it needs to be stopped at the same time as the core loop).

```
  void $ Async.wait logger
```
The logger is again stopped by the `Core`.

```
  Async.cancel metrics
```
The metrics server would run indefinitely, but since the core functionality of the application is done here already, we have to stop it as well.

```
  Async.cancel sync
```
The sync thread only has a one-way channel of communicating with the core, therefore it doesn't get a notification when the `Core` stops and therefore must be canceled explicitly.

server/src/Core.hs
fatho: Done

PR #24, opened 2017-10-11
Perform closing handshake when connection ends

fatho:


server/src/WebsocketServer.hs
```diff
@@ -104,9 +104,14 @@ handleClient conn path core = do
       currentValue <- getCurrentValue core path
       log (T.pack $ "Sending initial value to client: " ++ show currentValue) logRecords
       WS.sendTextData conn (Aeson.encode currentValue)
+
+    closeHandshake :: WS.ConnectionException -> IO ()
+    closeHandshake (WS.CloseRequest _ _) = WS.sendClose conn ("" :: Text)
```
rkrzr: I would add that the websocket protocol requires sending an empty frame as part of the closing handshake.

PR #26, opened 2017-10-12
Fix wrong values assigned to metrics

fatho:
Metric names are open for discussion.

Also did some logging cleanup.

### DIFF
--- a/server/src/Persistence.hs
+++ b/server/src/Persistence.hs
@@ -74,6 +74,9 @@ apply op val = do
 load :: PersistenceConfig -> IO (Either String PersistentValue)
 load config = runExceptT $ do
   value <- readData (pcDataFile config)
+  liftIO $ forM_ (pcMetrics config) $ \m -> do
+    size <- getFileSize (pcDataFile config)
+    Metrics.setDataSize size m
   valueVar <- lift $ newTVarIO value
   dirtyVar <- lift $ newTVarIO False
   journal <- for (pcJournalFile config) openJournal
@@ -110,7 +113,7 @@ sync val = do
     forM_ (pcMetrics . pvConfig $ val) $ \m -> do
       size <- getFileSize fileName
       Metrics.setDataSize size m
-      Metrics.setJournalSize size m
+      Metrics.setJournalSize (0 :: Int) m
       Metrics.incrementDataWritten size m
 
 -- * Private helper functions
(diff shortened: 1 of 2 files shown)


server/src/Metrics.hs
```diff
@@ -25,15 +25,15 @@ data IcepeakMetrics = IcepeakMetrics
...
                               "Size of journal file in bytes."))
-  <*> registerIO (counter (Info "icepeak_data_written" "Total number of bytes written so far."))
+  <*> registerIO (counter (Info "icepeak_data_written_bytes_total" "Total number of bytes written so far."))
   <*> registerIO (counter (Info "icepeak_journal_written_bytes_total"
                                 "Total number of bytes written to the journal so far."))
-  <*> registerIO (gauge (Info "icepeak_subscriber_count" "Number of websocket subscriber connections."))
+  <*> registerIO (gauge (Info "icepeak_subscribers" "Number of websocket subscriber connections."))
   where
-    requestCounter = counter (Info "icepeak_http_requests"
+    requestCounter = counter (Info "http_requests_total"
```
wesleybowman: why not `icepeak_http_requests_total`?

fatho: Citing the best practices

> ...should have a (single-word) application prefix relevant to the domain the metric belongs to. The prefix is sometimes referred to as namespace by client libraries. For metrics specific to an application, the prefix is usually the application name itself. Sometimes, however, metrics are more generic, like standardized metrics exported by client libraries. Examples:
>    - prometheus_notifications_total (specific to the Prometheus server)
>    - process_cpu_seconds_total (exported by many client libraries)
>    - http_request_duration_seconds (for all HTTP requests)

Especially the last example. Metrics with the same name can always be distinguished by the `node` label attached by prometheus.

wesleybowman: but this is specific to the icepeak server is it not?

fatho: I think the intention is that it helps when you want to aggregate data with other services. For example a graph of all HTTP requests that arrive anywhere, even across applications. That's easy to do if the metric has the same standardized name everywhere.

You can always separate the icepeak data by filtering on the `node` label.

fatho: But it seems they are not even consistent within their own style guide, a bit later, they have `api_http_requests_total` as an example. So I guess we can also leave the `icepeak_` prefix.
More important is, that the suffixes matches the unit and type (gauge/counter) of the metrics.

Issue #25, opened 2017-10-12
Allow requests with some sort of durability guarantee


fatho:
Currently, requests to the HTTP API post the command to the core queue and then immediately return. This could lead to a situation where a `GET` following a `PUT` returns stale data.

The proposal is to add a `durable` URL parameter that causes `PUT` and `DELETE` requests to wait until the modification has been applied to the data store. The easiest way is probably to create a new MVar in the request and pass it along the modification in the corresponding command:

https://github.com/channable/icepeak/blob/e6cfbec09c2e5dde0e6972702ade2ffdc99b3d04/server/src/Core.hs#L50

would become

```Haskell
 | Modify Modification (Maybe (MVar ()))
```

The `Nothing` case would behave as before, in the `Just` case, the `MVar` is filled after the modification has been processed and the request handler blocks with `takeMVar`.

ruuda:
Here’s a good overview of how to write data to disk in a way that ensures consistency: https://danluu.com/file-consistency/.

rkrzr:
We now support requests with a strong(er) durability guarantee. This was added in https://github.com/channable/icepeak/pull/27.

PR #27, opened 2017-10-12
Add URL option for waiting until request has been processed

fatho:


server/src/HttpServer.hs
```diff
@@ -34,14 +40,24 @@ new core =
...
       buildResponse result
 
+-- | Enqueue modification and wait for it to be processed, if desired by the client.
+postModification :: (Scotty.ScottyError e, MonadIO m) => Core -> Store.Modification -> Scotty.ActionT e m EnqueueResult
+postModification core op = do
+  -- the parameter is parsed as type (), therefore only presence or absence is important
+  durable <- maybeParam "durable"
+  waitVar <- liftIO $ for durable $ \() -> newEmptyMVar
+  result <- liftIO $ Core.tryEnqueueCommand (Core.Modify op waitVar) core
+  liftIO $ for_ waitVar $ void . takeMVar
```
rightfold: `void :: f () -> f ()` is redundant.

client-haskell/src/Icepeak/Client.hs
```diff
@@ -50,6 +54,25 @@ data Client = Client
...
+    -- request issued after the modification could still return stale data.
+    -- However, writes are ordered in that sending sequential (i.e. waiting on
+    -- the previous request to be answered) PUT/DELETE requests are processed in
+    -- order.
+  | DurabilityWait
+    -- ^ If the request is answered with a 202 response, the modification has
+    -- been applied and subsequent reads will receive the updated data.
+
+-- | Options to further specify request behavior
+data RequestOptions = RequestOptions { requestDurability :: Durability }
```
NunoAlexandre: This can be a `newtype` instead.

client-haskell/src/Icepeak/Client.hs
```diff
@@ -50,6 +54,25 @@ data Client = Client
...
 
+data Durability
+  = DurabilityEventual
+    -- ^ If the request is answered with a 202 response, it will eventually be
+    -- processed, but no guarantees are made about when that will happen. A GET
+    -- request issued after the modification could still return stale data.
+    -- However, writes are ordered in that sending sequential (i.e. waiting on
+    -- the previous request to be answered) PUT/DELETE requests are processed in
+    -- order.
+  | DurabilityWait
```
NunoAlexandre: `DurabilityWait` is a weird name...

client-haskell/src/Icepeak/Client.hs
```diff
@@ -50,6 +54,25 @@ data Client = Client
   , clientToken :: Maybe ByteString
   }
 
+data Durability
+  = DurabilityEventual
```
NunoAlexandre: Wouldn't `Effect` instead of `Durability` map better the idea involved here?

In any case, I prefer `EventualDurability` instead of `DurabilityEventual`.

server/src/Core.hs
```diff
@@ -47,11 +47,12 @@ import qualified Metrics
 data Command
   = Sync
     -- ^ The @Sync@ command causes the core to write the JSON value to disk.
-  | Modify Modification
+  | Modify Modification (Maybe (MVar ()))
     -- ^ The @Modify@ command applies a modification (writing or deleting) to the JSON value.
+    -- The optional MVar is used to signal that the command has been processed by the core.
   | Stop
     -- ^ The @Stop@ command causes the event loop of the Core to exit.
-  deriving (Eq, Show)
```
NunoAlexandre: Why removing Show?

fatho: Because you can't show an `MVar`

fatho:
This was an idea we had back then but it turned out things work fine without that durability guarantee, so it was never merged.

client-haskell/src/Icepeak/Client.hs
fatho: I don't like these names either, I just couldn't come up with something better at that time.

yoricksijsling:
Note that this is for issue #25.

Code all looks good to me, but how do I actually test this? How do i create a situation where passing the `durable` param will give a different result?

fatho:
@yoricksijsling You could spam it with updates to some key `/foo/a` and then a last update to `/foo/b`, then ask for the value of `/foo/b`. If processing all the updates takes long enough, it should sometimes return the old, sometimes the new value for `/foo/b` without the flag. But with that flag, it should always return the updated value.

client-haskell/src/Icepeak/Client.hs
rkrzr: Done.

client-haskell/src/Icepeak/Client.hs
rkrzr: I renamed it to `StrongDurability`.

client-haskell/src/Icepeak/Client.hs
rkrzr: I used `EventualDurability` now.

PR #31, opened 2018-01-05
Fix bug where exceptions did not get thrown immediately

### LINKED ISSUE #30: Socket-in-use exceptions are not thrown immediately
When i start `stack exec icepeak` while the port is in use, for instance by a `stack exec icepeak` in another terminal, i expected the application to throw an error immediately. However, it only throws this error after termination:

```
yorick@devm0:~/channable/icepeak/server$ stack exec icepeak
JWT authorization disabled.
Queue capacity is set to 256.
Sync: Persisting after every modification
System online. ** robot sounds **
Listening on port 3000.
^C
Termination sequence initiated ...
http server stopped with an exception: Network.Socket.bind: resource busy (Address already in use)
```

yoricksijsling:
Fixes #30. Exceptions in the http server or web socket server threads
would only get thrown after the three main
threads (core/http/websockets) have finished. Now the whole program
stops when one of the three throws an exception.

@fatho I'm cancelling the three main threads as well now if one of them throws an error. Is that actually safe to do or should they be stopped in a more friendly manner? (Especially the core thread)

### DIFF
--- a/server/app/Icepeak/Main.hs
+++ b/server/app/Icepeak/Main.hs
@@ -1,13 +1,11 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-module Main where
+module Main (main) where
 
-import Control.Concurrent.Async (Async)
-import Control.Exception (fromException, AsyncException)
+import Control.Exception (fromException, catch, AsyncException, SomeException)
 import Control.Monad (forM, void, when)
 import Data.Foldable (forM_)
 import Data.Semigroup ((<>))
-import Data.Text (Text)
 import Options.Applicative (execParser)
 import System.Environment (getEnvironment)
 import System.IO (BufferMode (..), hSetBuffering, stdout)
@@ -31,13 +29,12 @@ import qualified Metrics
 import qualified MetricsServer
 
 -- Install SIGTERM and SIGINT handlers to do a graceful exit.
-installHandlers :: Core -> Async () -> IO ()
-installHandlers core serverThread =
+installHandlers :: Core -> IO ()
+installHandlers core =
   let
     handle = do
       postLog (coreLogger core) "\nTermination sequence initiated ..."
       Core.postQuit core
-      Async.cancel serverThread
     handler = Signals.CatchOnce handle
     blockSignals = Nothing
     installHandler signal = Signals.installHandler signal handler blockSignals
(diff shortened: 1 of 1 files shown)


server/app/Icepeak/Main.hs
```diff
@@ -79,31 +79,52 @@ runCore core = do
...
-waitLog name logger action = Async.waitCatch action >>= handleLog where
-  handleLog (Left exc)
-    | Just (_ :: AsyncException) <- fromException exc = pure ()
-    | otherwise = Logger.postLog logger $ name <> " stopped with an exception: " <> Text.pack (show exc)
-  handleLog (Right _) = pure ()
+  runCoreResult <- waitFirstRunCoreResult pops upds serv
+  logRunCoreResult logger runCoreResult
+
+  -- kill all threads when one of the main threads ended
+  mapM_ Async.cancel [pops, upds, serv, metrics, sync]
```
fatho: What happens when one of the threads already ended? Is `cancel` a no-op then?

The core should be stopped with `Core.postQuit`, as that will cause it to save the current state to disk before stopping. Although the came could be achieved by catching exceptions inside the handler loop in the core.

But one could also argue that in cases of unhandled exceptions the data might be corrupted, and therefore should not be saved.

Thinking about this: I wonder what exceptions can actually happen here anyway, besides failing to open the socket. Maybe we should be more explicit about those errors and more thoughtful about the order in which we start up things.

server/app/Icepeak/Main.hs
```diff
@@ -79,31 +79,52 @@ runCore core = do
...
+-- | Data type to hold results for the async that finishes first
+data RunCoreResult
+  = RCR_CommandLoop SomeException
+  | RCR_WebSockets SomeException
+  | RCR_HTTP SomeException
+  | RCR_Success
+
+-- | Wait on the first async to finish
+waitFirstRunCoreResult :: Async () -> Async () -> Async () -> IO RunCoreResult
+waitFirstRunCoreResult x y z = waitAllThree' `finally` mapM_ cancel [x, y, z]
```
fatho: What does this do what `waitAnyCatchCancel` cannot do?

Maybe moving catching the exceptions into the async computations and having them return a status value is cleaner. But that's just an idea.
  
Also, the arguments seem to correspond to specific async actions (core, server, websockets), so they should be named accordingly.

yoricksijsling: Difference is that it allows us to see which of the threads threw the error. `waitAnyCatchCancel` only returns a `SomeException`, so they all look the same regardless of which thread threw it.

I've changed it now in the way you mentioned, so that the catch and tagging of the exception happens within the async. Now i can simply use `waitAny` instead of the, admittedly kind of odd, `waitFirstRunCoreResult`.

server/app/Icepeak/Main.hs
yoricksijsling: `cancel` is indeed a no-op if the thread already ended. I've changed the code to use `Core.postQuit` and wait for the core to stop.

server/app/Icepeak/Main.hs
```diff
@@ -77,33 +74,54 @@ runCore core = do
   httpServer <- HttpServer.new core
   let wsServer = WebsocketServer.acceptConnection core
 
+  let tryWithTag wrapper act = catch (act >> pure RCR_Fine) $ \exc -> case fromException exc of
```
rkrzr: Please add a comment to explain why we do this. Maybe lift the function to the top-level for that, then you can add a type signature as well.

server/app/Icepeak/Main.hs
```diff
@@ -77,33 +74,54 @@ runCore core = do
...
+        Just (_ :: AsyncException) -> pure $ RCR_Fine
+        _ -> pure $ wrapper exc
+
   -- start threads
-  pops <- Async.async $ Core.runCommandLoop core
+  pops <- Async.async $ tryWithTag RCR_CommandLoop $ Core.runCommandLoop core
+  upds <- Async.async $ tryWithTag RCR_WebSockets $ WebsocketServer.processUpdates core
+  serv <- Async.async $ tryWithTag RCR_HTTP $ Server.runServer logger wsServer httpServer
   sync <- Async.async $ Core.runSyncTimer core
```
rkrzr: Let's rename these, I just had a hard time remembering what `pops` stands for again..
E.g. `coreThread`, `websocketThread`, `httpThread`, `syncThread`. I think that's clearer.

server/app/Icepeak/Main.hs
```diff
@@ -77,33 +74,54 @@ runCore core = do
...
   Async.cancel sync
-
--- | Wait for an Async computation to exit and log unexpected exceptions.
-waitLog :: Text -> Logger -> Async () -> IO ()
-waitLog name logger action = Async.waitCatch action >>= handleLog where
-  handleLog (Left exc)
-    | Just (_ :: AsyncException) <- fromException exc = pure ()
-    | otherwise = Logger.postLog logger $ name <> " stopped with an exception: " <> Text.pack (show exc)
-  handleLog (Right _) = pure ()
+  void $ Async.wait pops
```
rkrzr: Maybe use `forM_` to kill all threads.

server/app/Icepeak/Main.hs
```diff
@@ -77,33 +74,54 @@ runCore core = do
...
+
+-- | Data type to hold results for the async that finishes first
```
rkrzr: Please add *why* we want to know this.

server/app/Icepeak/Main.hs
```diff
@@ -77,33 +74,54 @@ runCore core = do
...
+data RunCoreResult
+  = RCR_CommandLoop SomeException
+  | RCR_WebSockets SomeException
+  | RCR_HTTP SomeException
+  | RCR_Fine
```
rkrzr: We don't really use the prefix + _ convention.
I would perhaps rename this to:

```
data ThreadResult
  = CoreException SomeException
  |  WebsocketException SomeException
  |  HttpException SomeException
  | ThreadOk
```

server/app/Icepeak/Main.hs
yoricksijsling: Moved it to top level as:
```
-- | If a threads fails, catch the error and tag it so we know how to log it
catchRunCoreResult :: (SomeException -> RunCoreResult) -> IO () -> IO RunCoreResult
```

server/app/Icepeak/Main.hs
yoricksijsling: Much nicer indeed

server/app/Icepeak/Main.hs
yoricksijsling: Skipping this one because it doesn't really read better:
```
  mapM_ Async.cancel [ void webSocketThread
                     , void httpThread
                     , metricsThread
                     , syncThread
                     ]
```

PR #32, opened 2018-01-05
Make icepeak port configurable

### LINKED ISSUE #12: Make port configurable
Currently, Icepeak always listens on port 3000. We should make that configurable with a command-line option and possibly also via an environment variable.

yoricksijsling:
Fixes #12. Port is configurable through --port and $ICEPEAK_PORT.

Added a test for it:

```
icepeak/server$ integration-tests/using-different-ports.sh 

## Should default on port 3000
PASSED

## Should use --port if given
PASSED

## Should use ICEPEAK_PORT if given
PASSED

## Should prefer --port over ICEPEAK_PORT
PASSED

## Should ignore ICEPEAK_PORT if it is not a number
PASSED

## Should fail if --port is not a number
PASSED

ALL PASSED
```

### DIFF
--- a/server/src/Config.hs
+++ b/server/src/Config.hs
@@ -20,6 +20,7 @@ import qualified Web.JWT as JWT
 -- command-line arguments
 data Config = Config
   { configDataFile :: FilePath
+  , configPort :: Int
     -- | Enables the use of JWT for authorization in JWT.
   , configEnableJwtAuth :: Bool
     -- | The secret used for verifying the JWT signatures. If no secret is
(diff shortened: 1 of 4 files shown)


server/src/Config.hs
```diff
@@ -79,6 +84,12 @@ configParser environment = Config
              help "Enable journaling. This only has an effect when periodic syncing is enabled.")
   where
     environ var = foldMap value (lookup var environment)
+
+    readFromEnvironment :: Read a => String -> Maybe a
+    readFromEnvironment var = case reads <$> lookup var environment of
```
fatho: You can use [`Text.Read.readMaybe`](https://hackage.haskell.org/package/base-4.10.1.0/docs/Text-Read.html#v:readMaybe).

server/src/Server.hs
```diff
@@ -5,6 +5,9 @@ module Server
 )
 where
 
+import Data.Semigroup      ((<>))
```
fatho: Please remove the superfluous spaces. We generally have our imports alphabetically sorted and without any alignment in order to prevent diffs consisting mostly of realignments.

yoricksijsling:
Should be fine now @fatho and @rkrzr

PR #34, opened 2018-03-09
Create an empty database if there is none yet

rkrzr:
If we cannot find a database at the configured file path we used to
crash with a not terribly good error message.

This PR instead logs a warning and then creates an empty database
(an empty JSON object: {}). Note, that the object will only be
persisted when some changes are written to disk.

### DIFF
--- a/server/src/Persistence.hs
+++ b/server/src/Persistence.hs
@@ -176,7 +176,7 @@ recoverJournal pval = for_ (pvJournal pval) $ \journalHandle -> ExceptT $ fmap f
 -- | Read and decode the data file from disk
 readData :: FilePath -> Logger -> ExceptT String IO Store.Value
 readData filePath logger = ExceptT $ do
-  eitherEncodedValue <- tryIOError $ withFile filePath ReadMode SBS.hGetContents
+  eitherEncodedValue <- tryIOError $ SBS.readFile filePath
   case eitherEncodedValue of
     Left e | isDoesNotExistError e -> do
         -- If there is no icepeak.json file yet, we create an empty one instead.
--- a/server/src/Persistence.hs
+++ b/server/src/Persistence.hs
@@ -73,7 +73,7 @@ apply op val = do
 -- | Load the persisted data from disk and recover journal entries.
 load :: PersistenceConfig -> IO (Either String PersistentValue)
 load config = runExceptT $ do
-  value <- readData (pcDataFile config)
+  value <- readData (pcDataFile config) (pcLogger config)
   liftIO $ forM_ (pcMetrics config) $ \m -> do
     size <- getFileSize (pcDataFile config)
     Metrics.setDataSize size m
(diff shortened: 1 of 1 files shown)


server/src/Persistence.hs
```diff
@@ -173,12 +173,19 @@ recoverJournal pval = for_ (pvJournal pval) $ \journalHandle -> ExceptT $ fmap f
...
--- | Read and decode the data file
-readData :: FilePath -> ExceptT String IO Store.Value
-readData filePath = ExceptT $ do
+-- | Read and decode the data file from disk
+readData :: FilePath -> Logger -> ExceptT String IO Store.Value
+readData filePath logger = ExceptT $ do
   eitherEncodedValue <- try $ withFile filePath ReadMode SBS.hGetContents
   case (eitherEncodedValue :: Either SomeException SBS.ByteString) of
-    Left exc -> pure $ Left $ "Failed to read the data from disk: " ++ show exc
+    Left _exc -> do
```
fatho: This should check whether the exception was indeed that the file was not found, and only in that case it should attempt to create it.

server/src/Persistence.hs
```diff
@@ -173,12 +173,19 @@ recoverJournal pval = for_ (pvJournal pval) $ \journalHandle -> ExceptT $ fmap f
...
+      -- If there is no icepeak.json file yet, we create an empty one instead.
+      let emptyObject = Aeson.object []
+          message = "WARNING: Failed to read the data from disk. " <>
+                    "Created an empty database instead."
+      Logger.postLogBlocking logger message
+      pure $ Right emptyObject
```
fatho: It should also immediately try to write the file, because if the location is not accessible at all, and it fails to save only after the DB contains data that should be persisted, data will be lost.

ruuda:
Can we merge this? It is a papercut for new developers.

server/src/Persistence.hs
rkrzr: Good point, fixed.

server/src/Persistence.hs
rkrzr: True, fixed.

server/src/Persistence.hs
```diff
@@ -173,12 +173,29 @@ recoverJournal pval = for_ (pvJournal pval) $ \journalHandle -> ExceptT $ fmap f
...
--- | Read and decode the data file
-readData :: FilePath -> ExceptT String IO Store.Value
-readData filePath = ExceptT $ do
-  eitherEncodedValue <- try $ withFile filePath ReadMode SBS.hGetContents
-  case (eitherEncodedValue :: Either SomeException SBS.ByteString) of
-    Left exc -> pure $ Left $ "Failed to read the data from disk: " ++ show exc
+-- | Read and decode the data file from disk
+readData :: FilePath -> Logger -> ExceptT String IO Store.Value
+readData filePath logger = ExceptT $ do
+  eitherEncodedValue <- tryIOError $ withFile filePath ReadMode SBS.hGetContents
```
fatho: It should be possible to replace `withFile filePath ReadMode SBS.hGetContents` with `SBS.readFile filePath`. Don't know why we (I?) made this so complicated back then.

rkrzr: Yes, fixed.

wesleybowman:
> was getting some tea and got side-tracked for a little while

:joy:

PR #37, opened 2018-08-17
Upgrade to stack lts-12.6 (from lts-11.20)

rkrzr:
This also upgrades GHC from version 8.2 to 8.4.

There were some breaking changes in the `prometheus-client` package.
They changed their API in various places. The changes are described in their changelog here:

https://github.com/fimad/prometheus-haskell/blob/master/prometheus-client/CHANGELOG.md

I adjusted our Metrics.hs module to work with the new API.
For details, see the commit messages.

### DIFF
--- a/server/src/Metrics.hs
+++ b/server/src/Metrics.hs
@@ -1,7 +1,6 @@
 {-# LANGUAGE OverloadedStrings #-}
 module Metrics where
 
-import Control.Monad (void)
 import Data.Text (Text, pack)
 import Data.Text.Encoding (decodeUtf8With)
 import Data.Text.Encoding.Error (lenientDecode)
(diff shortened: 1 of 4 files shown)


server/src/Metrics.hs
```diff
@@ -1,58 +1,68 @@
...
+import Data.Text (Text, pack)
+import Data.Text.Encoding (decodeUtf8With)
+import Data.Text.Encoding.Error (lenientDecode)
 
-type HttpMethodLabel = String
-type HttpStatusLabel = String
+import qualified Network.HTTP.Types as Http
 
-type HttpRequestCounter = Vector (HttpMethodLabel, HttpStatusLabel) Counter
+import Prometheus
```
ruuda: I would either make this a qualified import, or explicitly import the functions and types used from this module.

server/src/Persistence.hs
```diff
@@ -61,7 +60,7 @@ apply op val = do
     LBS8.hPutStrLn journalHandle entry
     for_ (pcMetrics . pvConfig $ val) $ \metrics -> do
       journalPos <- hTell journalHandle
-      Metrics.incrementJournalWritten (LBS8.length entry) metrics
+      _ <- Metrics.incrementJournalWritten (LBS8.length entry) metrics
```
ruuda: Can also be written as `void $ Metrics.increment...`

Why does it return a `Bool` though, what does it mean?

server/stack.yaml
```diff
@@ -1 +1,6 @@
-resolver: lts-11.20
+resolver: lts-12.6
+
+extra-deps:
+  - prometheus-client-1.0.0
```
ruuda: So the [`prometheus`](https://github.com/bitnomial/prometheus) package is now included in Stackage after I asked for it to be included, but we use a different Prometheus library here? We do use `prometheus` in one of our other projects. Maybe we should ask the author of `prometheus-client` to include it in Stackage too then? Would be nice if we can reduce the number of extra deps.

server/src/Persistence.hs
fatho: Apparently it returns whether it actually increased the counter. When the argument is negative, it won't change the counter and return `False`, otherwise, it'll add the value and return `True`.

server/stack.yaml
rkrzr: I just checked and we use the `prometheus-client` library in WM and SHM, and we use the `prometheus` library in JM. I don't know the pros and cons of either (the former has 36 GH stars and the latter 19 fwiw).
I'll just leave it like it is for now, since it's working well enough, but we can discuss later to unify on one of those libraries.

server/src/Metrics.hs
rkrzr: True, I changed it to explicit imports now. We should do that everywhere in IP eventually.

server/src/Persistence.hs
rkrzr: I added a comment.

PR #38, opened 2018-10-09
Pair HTTP manager with client details in one type

rightfold:
In most use cases, you want to propagate these two together anyway. It is still
possible not to, if you want.

This is a breaking change, but we use commit-pinned versions so it's not an issue for us. People who use master as their reference will have to change their code.

### DIFF
--- a/client-haskell/src/Icepeak/Client.hs
+++ b/client-haskell/src/Icepeak/Client.hs
@@ -3,17 +3,19 @@
 -- | Functions are exported in descending order of abstraction level.
 --
 -- >>> :set -XOverloadedStrings
--- >>> import Icepeak.Client (Client (..), deleteAtLeaf, setAtLeaf)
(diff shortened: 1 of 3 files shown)


client-haskell/src/Icepeak/Client.hs
```diff
@@ -47,12 +49,15 @@ import qualified Network.HTTP.Types.Status as HTTP
...
-data Client = Client
-  { clientHost  :: ByteString
-  , clientPort  :: Word16
-  , clientToken :: Maybe ByteString
-  }
+-- | Details paired with HTTP manager for making requests.
+data Client = Client HTTP.Manager Details
+
+-- | Client connection details: how to connect to Icepeak?
+data Details = Details
```
ruuda: Nit: `Details` is a very generic name, something like `ConnectionSettings` would be more descriptive.

rightfold: Renamed to Config.

PR #46, opened 2019-04-26
Set up CI with Azure Pipelines

rkrzr:


rkrzr:
This was a Hackathon project to evaluate Azure Pipelines, and we will not be moving forward with this. Closing.

PR #49, opened 2019-08-10
Add client retries

### LINKED ISSUE #44: Add retries to the Icepeak client
The icepeak client does currently not support retrying. We worked around that in JM by adding retries ourselves, but it would be nicer and more logical if the client could do that itself.

This could be implemented by adding a `RetryPolicy` field to the client's config. This could be an optional field, which just defaults to no retries, when not specified by the user.

The new `Config` type could for example look like this:

```haskell
import Conrol.Retry (RetryPolicy)

data Config = Config
  }
  { configHost  :: ByteString
  , configPort  :: Word16
  , configToken :: Maybe ByteString
  , configRetryPolicy :: Maybe RetryPolicy  -- new field
  } deriving (Eq, Show)
```

DStranger:
Fixes #44

This is my attempt at adding retries to the Icepeak client.
I'm not really sure how to test this though, maybe you have suggestions?

### DIFF
--- a/client-haskell/src/Icepeak/Client.hs
+++ b/client-haskell/src/Icepeak/Client.hs
@@ -9,7 +9,7 @@
 -- >>> import Control.Retry (constantDelay, limitRetries)
 -- >>> httpManager <- HTTP.newManager HTTP.defaultManagerSettings
 -- >>> let retryPolicy = constantDelay 50000 <> limitRetries 5
--- >>> let client = Client httpManager (Config "localhost" 3000 (Just "<token>")) (Just retryPolicy)
+-- >>> let client = Client httpManager (Config "localhost" 3000 (Just "<token>")) retryPolicy
 -- >>> setAtLeaf client ["foo", "bar", "baz"] ([Just 1, Just 2, Nothing, Just 4] :: [Maybe Int])
 -- Status {statusCode = 202, statusMessage = "Accepted"}
 -- >>> deleteAtLeaf client ["foo", "bar"]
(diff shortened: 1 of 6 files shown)


client-haskell/src/Icepeak/Client.hs
```diff
@@ -48,9 +53,14 @@ import qualified Network.HTTP.Client as HTTP
...
 import qualified Network.HTTP.Types.Header as Header
 import qualified Network.HTTP.Types.URI as URI
+import qualified Control.Retry as Retry
 
 -- | Config paired with HTTP manager for making requests.
-data Client = Client HTTP.Manager Config
+data Client m = Client 
+  { clientHttpManager :: HTTP.Manager
+  , clientConfig      :: Config
+  , clientRetryPolicy :: Maybe (RetryPolicyM m)
```
rightfold: Nothing and Just doNotRetry seem to encode the same information. I suggest making this field not Maybe, and exporting doNotRetry. That would simplify both the implementation and the usage of the library.

rightfold: Adding a new field here breaks the integration tests:

https://github.com/channable/icepeak/blob/fb30e95429010f1d46a29ca312724437da5ac596/server/integration-tests/stress_test.hs#L36

https://github.com/channable/icepeak/blob/fb30e95429010f1d46a29ca312724437da5ac596/server/integration-tests/auth_stress_test.hs#L39

DStranger: Yes, this seems like a good idea!

DStranger: Indeed, missed this.

DStranger:
> I got two remarks but other than that: looks good! Very nice. 👍
Thanks!
I fixed those two things, had to force push after a rebase on master.

PR #51, opened 2019-09-09
Document durable requests

joris-burgers:
Document durable requests in the IcePeak protocol.

### DIFF
--- a/docs/IcepeakProtocol.md
+++ b/docs/IcepeakProtocol.md
@@ -8,6 +8,16 @@ establishes the connection to the server.
 
 An example URL is `ws://yourdomain.com/path/to/channel`.
 
+### Receiving data
+
+There are 2 ways to receive data from an Icepeak server. The first is with a direct GET command over
+the HTTP protocol. An example of such a call would be `GET http://localhost:3000/path/to/channel`. 
+This would result in the information that is stored under the `path/to/channel` object, if any. 
+The result would be a JSON object with all keys. 
+
+The other way to receive data is with a GET call that is upgraded to a WebSocket.
+As long as the connection remains open, Icepeak will send updates whenever a value changes on
+the server. For example, the connection `ws://localhost:3000/path/to/channel`. 
 `/path/to/channel` is then the identifier of a JSON collection.
 If any value of that collection changes, the whole collection under this path will be sent to
 the client.
@@ -19,3 +29,10 @@ Example:
 3. The server immediately sends the current value of `/path/to/channel`, e.g. `{"status": "updating"}`
 4. The value gets updated to `{"status": "done"}` (by another process)
 5. The server then immediately sends the updated value to all clients listening to `/path/to/channel`
+
+### Modifying data
+
+Data on Icepeak can be modified by either a PUT request or a DELETE request. The PUT will introduce
+or overwrite new information to the Icepeak server and DELETE will remove information from the server.
+In case some information changes, all information that is changed will be send to every 
+client that is listening to that channel. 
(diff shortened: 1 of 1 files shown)


docs/IcepeakProtocol.md
```diff
@@ -19,3 +29,42 @@ Example:
 3. The server immediately sends the current value of `/path/to/channel`, e.g. `{"status": "updating"}`
 4. The value gets updated to `{"status": "done"}` (by another process)
 5. The server then immediately sends the updated value to all clients listening to `/path/to/channel`
+
+### Modifying data
+
+Data on Icepeak can be modified by either a PUT request or a DELETE request. The PUT will introduce
+or overwrite new information to the Icepeak server and DELETE will remove information from the server.
```
fatho: [...] introduce new or overwrite existing [...]

docs/IcepeakProtocol.md
```diff
@@ -19,3 +29,42 @@ Example:
...
+  - If a value along the path is already an object, the field specified by the
+    path is added to it.
+
+- `Delete` removes a JSON value at a given path.
+
+  - If the path does not exist, nothing happens.
+  - If the path refers to a field in an object, that field is removed from the
+    object.
+
+Both of these operations have an optional parameter, called `durable`, which if set,
```
fatho: What does being "set" mean in this context. It could be
- `foo/bar/baz?durable`
- `foo/bar/baz?durable=true`
- `foo/bar/baz?durable=yes`
- `foo/bar/baz?durable=1`
- etc.

Please specify which of these is meant.

docs/IcepeakProtocol.md
```diff
@@ -19,3 +29,42 @@ Example:
...
+ensures that the action is completed before the response code 202 (accepted) is
+send back to the client. This ensures that no stale data is possible to get via
+a get call. 
+
+Example if `durable` is **not** set:
+1. The client sends a PUT command to the server with a new value for `foo`. 
+2. The server responds with a 202 (accepted) response immidiatly, before completing the action.
+3. A GET request is send to the server, requesting the value for `foo`, but `foo` is not yet updated on the server.
```
fatho: This can always happen if the GET request is sent from another server, as a GET request will not go through the Core event loop.
For the purpose of demonstrating the difference, it will be more interesting to consider a GET request that has been sent from the same client that also sent the PUT command.

docs/IcepeakProtocol.md
```diff
@@ -19,3 +29,42 @@ Example:
...
+4. The server responds with an out-of-date copy of the data. 
+5. The update of `foo` is executed on the server and the result is broadcasted.
+
+Example if durable is set:
+1. The client sends a PUT command to the server with a new value for `foo`. 
+2. The server waits for the new value of `foo` to be saved on the server.
+3. The server responds with a 202 (accepted) response.
+4. A GET request is send to the server, requesting the value for `foo`.
```
fatho: As stated above, you should specify that the GET request comes from the same client for the purpose of this demonstration. From the clients perspective, the flow is:

1. send `PUT` request to server
2. wait for response to `PUT` request
3. send `GET` request

The difference between durable and non-durable will then be observable in the response to the final `GET` request.

docs/IcepeakProtocol.md
fatho: And actually, even the latter is not guaranteed if multiple concurrent users send a PUT request to the same path on Icepeak with different values.

So a durable request doesn't really guarantee that you `GET` the same value that you `PUT` before, but rather that if the `PUT` response has been received, the client can be sure it's update has been processed.
Conversely, it could very well be that the request was processed, but that the client didn't receive the response (e.g. due to a network issue).

PR #52, opened 2019-09-10
Add sentry logging to Icepeak

### LINKED ISSUE #43: Add Sentry logging for fatal crashes
If any one of our control threads receives an unhandled exception we also terminate all other threads and cleanly shut down the service after handling all remaining commands in the command queue (unless there is a bug like in https://github.com/channable/icepeak/issues/40).
We then rely on systemd to restart the service and to resume operation.

It would be good if we would also be alerted of these unexpected exceptions.
We should add a Sentry handler that logs these exceptions to Sentry to make it easy to spot and fix any unexpected issues.
This would also have made debugging in https://github.com/channable/icepeak/issues/40 easier.

joris-burgers:
Fixed #43

### DIFF
--- a/server/app/Icepeak/Main.hs
+++ b/server/app/Icepeak/Main.hs
@@ -5,6 +5,7 @@ module Main (main) where
 import Control.Exception (fromException, catch, AsyncException, SomeException)
 import Control.Monad (forM, void, when)
 import Data.Foldable (forM_)
+import Data.Maybe (fromMaybe)
 import Data.Semigroup ((<>))
 import Options.Applicative (execParser)
 import System.Environment (getEnvironment)
@@ -15,6 +16,7 @@ import qualified Data.Text as Text
 import qualified Prometheus
 import qualified Prometheus.Metric.GHC
 import qualified System.Posix.Signals as Signals
+import qualified System.Log.Raven.Types as Sentry
 
 import Config (Config (..), configInfo)
 import Core (Core (..))
@@ -27,6 +29,7 @@ import qualified WebsocketServer
 import qualified Logger
 import qualified Metrics
 import qualified MetricsServer
+import qualified CrashLogging
 
 -- Install SIGTERM and SIGINT handlers to do a graceful exit.
 installHandlers :: Core -> IO ()
@@ -45,33 +48,35 @@ installHandlers core =
 
 main :: IO ()
 main = do
-  -- make sure output is flushed regularly
-  hSetBuffering stdout LineBuffering
+  crashLogger <- CrashLogging.getCrashLogger
+  maybe id (CrashLogging.runWithCrashLogger "Main") crashLogger $ do
+    -- make sure output is flushed regularly
+    hSetBuffering stdout LineBuffering
+    --_ <- error "Some error occured"
+    env <- getEnvironment
+    config <- execParser (configInfo env)
 
-  env <- getEnvironment
-  config <- execParser (configInfo env)
+    -- start logging as early as possible
+    logger <- Logger.newLogger (fromIntegral $ configQueueCapacity config)
+    loggerThread <- Async.async $ Logger.processLogRecords logger
 
-  -- start logging as early as possible
-  logger <- Logger.newLogger (fromIntegral $ configQueueCapacity config)
-  loggerThread <- Async.async $ Logger.processLogRecords logger
+    -- setup metrics if enabled
+    icepeakMetrics <- forM (configMetricsEndpoint config) $ const $ do
+      void $ Prometheus.register Prometheus.Metric.GHC.ghcMetrics
+      Metrics.createAndRegisterIcepeakMetrics
 
-  -- setup metrics if enabled
-  icepeakMetrics <- forM (configMetricsEndpoint config) $ const $ do
-    void $ Prometheus.register Prometheus.Metric.GHC.ghcMetrics
-    Metrics.createAndRegisterIcepeakMetrics
+    eitherCore <- Core.newCore config logger icepeakMetrics
+    either putStrLn (runCore crashLogger) eitherCore
 
-  eitherCore <- Core.newCore config logger icepeakMetrics
-  either putStrLn runCore eitherCore
+    -- only stop logging when everything else has stopped
+    Logger.postStop logger
+    Async.wait loggerThread
 
-  -- only stop logging when everything else has stopped
-  Logger.postStop logger
-  Async.wait loggerThread
-
-runCore :: Core -> IO ()
-runCore core = do
+runCore :: Maybe Sentry.SentryService -> Core -> IO ()
+runCore mSentryService core = do
   let config = coreConfig core
   let logger = coreLogger core
-  httpServer <- HttpServer.new core
+  httpServer <- HttpServer.new mSentryService core
   let wsServer = WebsocketServer.acceptConnection core
 
   -- start threads
--- a/server/src/HttpServer.hs
+++ b/server/src/HttpServer.hs
@@ -2,6 +2,7 @@
 
 module HttpServer (new) where
 
+import Control.Exception
 import Control.Concurrent.MVar (newEmptyMVar, takeMVar)
 import Control.Monad
 import Control.Monad.IO.Class
@@ -15,26 +16,29 @@ import qualified Data.Text.Lazy as LText
 import qualified Network.Wai as Wai
 import qualified Web.Scotty.Trans as Scotty
 
+import qualified System.Log.Raven.Types as Sentry
+
 import JwtMiddleware (jwtMiddleware)
 import Core (Core (..), EnqueueResult (..))
 import Config (Config (..))
 import qualified Store
 import qualified Core
 import qualified Metrics
+import qualified CrashLogging
 
-new :: Core -> IO Application
-new core =
+new :: Maybe Sentry.SentryService -> Core -> IO Application
+new mSentryService core =
   scottyApp $ do
     -- first middleware is the outermost. this has to be the metrics middleware
     -- in order to intercept all requests their corresponding responses
     forM_ (coreMetrics core) $ middleware . metricsMiddleware
-
+    Scotty.defaultHandler (\e -> liftIO $ CrashLogging.logCrashMessage "Handler" mSentryService (show e) >> error (show e))
     when (configEnableJwtAuth $ coreConfig core) $
       middleware $ jwtMiddleware $ configJwtSecret $ coreConfig core
 
     get (regex "^") $ do
       path <- Wai.pathInfo <$> request
-      maybeValue <- liftIO $ Core.getCurrentValue core path
+      maybeValue <- Scotty.liftAndCatchIO $ Core.getCurrentValue core path
       maybe (status status404) json maybeValue
 
     put (regex "^") $ do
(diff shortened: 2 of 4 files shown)


server/app/Icepeak/Main.hs
```diff
@@ -123,7 +127,9 @@ logRunCoreResult logger rcr = do
   where
     handleLog name exc
       | Just (_ :: AsyncException) <- fromException exc = pure ()
-      | otherwise = Logger.postLog logger $ name <> " stopped with an exception: " <> Text.pack (show exc)
+      | otherwise = do
+            Logger.postLog logger $ name <> " stopped with an exception: " <> Text.pack (show exc)
+            SentryLogging.logException "Log Run Core Error" mSentryService exc
```
fatho: The error that is logged to sentry should also contain the `name` of the thread that crashed.

server/src/SentryLogging.hs
```diff
@@ -0,0 +1,47 @@
...
+import qualified System.Log.Raven.Transport.HttpConduit as Sentry
+import qualified System.Log.Raven.Types as Sentry
+
+-- | Returns a Maybe SentryService which can be used to send error information
+-- to Sentry. Return value is Nothing is the environment variable SENTRY_DSN is
+-- not set.
+getCrashLogger :: IO (Maybe Sentry.SentryService)
+getCrashLogger = do
+    let
+    maybeSentryDSN <- lookupEnv "SENTRY_DSN"
```
fatho: Getting the Sentry DSN should be part of the Config module, so that we have all configuration-related stuff together in one place. That would also allow specifying it as command line argument, if desired.
You can have a look at how the `JWT_SECRET` is handled in the config.

server/src/HttpServer.hs
```diff
@@ -15,26 +16,35 @@ import qualified Data.Text.Lazy as LText
...
     -- in order to intercept all requests their corresponding responses
     forM_ (coreMetrics core) $ middleware . metricsMiddleware
-
+    -- Use the Sentry logger if available
+    -- Scottys error handler will only catch errors that are thrown from within
+    -- a ```liftAndCatchIO``` function.
+    if isJust mSentryService then
+        Scotty.defaultHandler (liftIO . SentryLogging.logCrashMessage "Thread handler error" mSentryService . show)
+    else
+        Scotty.defaultHandler (liftIO . putStrLn . LText.unpack )
```
fatho: Instead of printing directly to stdout, it should go via the global logger (which you can get via `coreLogger core`).
That way, we can control what to do with log messages in one place.

server/app/Icepeak/Main.hs
```diff
@@ -45,33 +47,35 @@ installHandlers core =
...
-    Metrics.createAndRegisterIcepeakMetrics
+    -- setup metrics if enabled
+    icepeakMetrics <- forM (configMetricsEndpoint config) $ const $ do
+      void $ Prometheus.register Prometheus.Metric.GHC.ghcMetrics
+      Metrics.createAndRegisterIcepeakMetrics
 
-  eitherCore <- Core.newCore config logger icepeakMetrics
-  either putStrLn runCore eitherCore
+    eitherCore <- Core.newCore config logger icepeakMetrics
+    either putStrLn (runCore crashLogger) eitherCore
```
fatho: This `putStrLn` is probably an oversight and should also go via the logger. Could you fix it as well?

server/app/Icepeak/Main.hs
```diff
@@ -47,35 +46,33 @@ installHandlers core =
...
+  loggerThread <- Async.async $ Logger.processLogRecords logger
+  SentryLogging.runWithCrashLogger "Control loop error" (Logger.loggerSentryService logger) $ do
     -- setup metrics if enabled
     icepeakMetrics <- forM (configMetricsEndpoint config) $ const $ do
       void $ Prometheus.register Prometheus.Metric.GHC.ghcMetrics
       Metrics.createAndRegisterIcepeakMetrics
 
     eitherCore <- Core.newCore config logger icepeakMetrics
-    either putStrLn (runCore crashLogger) eitherCore
+    either (postLog logger LogInfo . Text.pack) runCore eitherCore
```
fatho: This should be a `LogError`. `eitherCore` will only be `Left` when something went wrong during initialization.

server/src/HttpServer.hs
```diff
@@ -16,29 +16,25 @@ import qualified Data.Text.Lazy as LText
...
     -- in order to intercept all requests their corresponding responses
     forM_ (coreMetrics core) $ middleware . metricsMiddleware
     -- Use the Sentry logger if available
     -- Scottys error handler will only catch errors that are thrown from within
     -- a ```liftAndCatchIO``` function.
-    if isJust mSentryService then
-        Scotty.defaultHandler (liftIO . SentryLogging.logCrashMessage "Thread handler error" mSentryService . show)
-    else
-        Scotty.defaultHandler (liftIO . putStrLn . LText.unpack )
+    Scotty.defaultHandler (liftIO . postLog (coreLogger core) LogError . pack . show)
```
fatho: If I understand the documentation correctly, the default handler is also responsible for actually responding the HTTP request.
That can be a simple `503` with a plain text "Internal server error" body.

server/src/HttpServer.hs
```diff
@@ -16,29 +16,25 @@ import qualified Data.Text.Lazy as LText
 import qualified Network.Wai as Wai
 import qualified Web.Scotty.Trans as Scotty
 
-import qualified System.Log.Raven.Types as Sentry
-
 import JwtMiddleware (jwtMiddleware)
 import Core (Core (..), EnqueueResult (..))
 import Config (Config (..))
+import Logger
```
fatho: We only use wildcard imports in exceptional circumstances. You should use either use an unqualified import with an explicit import list, or a qualified import.
The reason for this is that otherwise it is quite hard to find out where a given symbol comes from.

server/src/Logger.hs
```diff
@@ -3,46 +3,63 @@ module Logger
...
   deriving (Eq, Ord, Show, Read)
 
-newLogger :: Natural -> IO Logger
-newLogger queueSize = Logger <$> atomically (newTBQueue queueSize)
+newLogger :: Config -> IO Logger
+newLogger config = Logger
+  <$> atomically (newTBQueue (fromIntegral $ configQueueCapacity config))
+  <*> if configDisableSentryLogging config then
+        pure Nothing
+      else maybe (pure Nothing) (fmap Just . getCrashLogger) (configSentryDSN config)
```
fatho: This would be cleaner with a where clause, i.e.

```haskell
newLogger config = Logger
  <$> createQueue
  <*> createSentryService
  where
    createQueue = ...
    createSentryService = ...
```

server/src/Logger.hs
```diff
@@ -3,46 +3,63 @@ module Logger
...
 import Control.Concurrent.STM (atomically)
 import Control.Concurrent.STM.TBQueue (TBQueue, newTBQueue, readTBQueue, writeTBQueue, isFullTBQueue)
-import Data.Text (Text)
-import GHC.Natural (Natural)
+import Data.Text (Text, unpack)
+import Data.Maybe (isJust)
 import Prelude hiding (log)
 
+
+
```
fatho: One empty line between qualified and unqualified imports is enough.

server/src/Logger.hs
```diff
@@ -3,46 +3,63 @@ module Logger
...
   postStop,
-  processLogRecords
+  processLogRecords,
+  loggerSentryService
 )
 where
 
-import Control.Monad (unless)
+import SentryLogging
+import Config
```
fatho: Same thing about wildcard imports.

server/app/Icepeak/Main.hs
```diff
@@ -52,20 +53,20 @@ main = do
   config <- execParser (configInfo env)
 
   -- start logging as early as possible
-  logger <- Logger.newLogger (fromIntegral $ configQueueCapacity config)
+  logger <- Logger.newLogger config
   loggerThread <- Async.async $ Logger.processLogRecords logger
+  SentryLogging.runWithCrashLogger "Control loop error" (Logger.loggerSentryService logger) $ do
```
fatho: All errors caught by this would now go only via sentry and not be logged wherever else.
This handler should probably also just post the log message to the logger.

server/app/Icepeak/Main.hs
```diff
@@ -52,20 +53,20 @@ main = do
...
 
-  eitherCore <- Core.newCore config logger icepeakMetrics
-  either putStrLn runCore eitherCore
-
-  -- only stop logging when everything else has stopped
-  Logger.postStop logger
-  Async.wait loggerThread
+    -- only stop logging when everything else has stopped
+    Logger.postStop logger
+    Async.wait loggerThread
```
fatho: Given my remark above, stopping the logger should then be outside of the scope of the handler above, so that we don't end up in a situation where that unexpectedly fails and then tries to post something to the logger that has already been stopped.

server/src/Config.hs
```diff
@@ -32,6 +32,8 @@ data Config = Config
   , configSyncIntervalMicroSeconds :: Maybe Int
   -- | Enable journaling, only in conjunction with periodic syncing
   , configEnableJournaling :: Bool
+  , configDisableSentryLogging :: Bool
+  , configSentryDSN :: Maybe String
```
fatho: Could you add some haddock comments here explaining the relation of the two, and the reason why they exist?
I suppose the `configDisableSentryLogging` is useful for the case where a `SENTRY_DSN` environment variable is set, but we actually don't want to log to sentry.

server/src/SentryLogging.hs
```diff
@@ -0,0 +1,40 @@
...
+getCrashLogger dsn = Sentry.initRaven dsn id Sentry.dumpRecord Sentry.errorFallback
+
+-- | Send some exception that has occured to Sentry. Function does nothing when
+-- Sentry service is Nothing
+logException :: String -> Maybe Sentry.SentryService -> SomeException -> IO ()
+logException name mSentryService exception =
+  maybe
+    (pure ())
+    (\ss -> Sentry.register ss name Sentry.Fatal (show (exception :: SomeException)) id)
+    mSentryService
```
fatho: I would suggest writing it as
```haskell
for_ mSentryService $ \service ->
  Sentry.register service name Sentry.Fatal (show (exception :: SomeException)) id
```
Using `for_` from `Data.Foldable`.

server/src/Logger.hs
```diff
@@ -53,8 +70,12 @@ processLogRecords logger = go
     go = do
       cmd <- atomically $ readTBQueue (loggerQueue logger)
       case cmd of
-        LogRecord logRecord -> do
+        LogRecord logLevel logRecord -> do
           T.putStrLn logRecord
+          when (logLevel == LogError && isJust (loggerSentryService logger) ) (
+              putStrLn "Sending error message to Sentry" >>
```
fatho: I don't think we need this in production. If sending succeeds, it was redundant, if sending fails, we'll get the output of the fallback logger configured in the sentry service.

server/src/Logger.hs
```diff
@@ -43,10 +41,13 @@ data LogCommand = LogRecord LogLevel LogRecord | LogStop
...
-  <*> if configDisableSentryLogging config then
-        pure Nothing
-      else maybe (pure Nothing) (fmap Just . getCrashLogger) (configSentryDSN config)
+  <$> createQueue
+  <*> createSentryService
+  where
+    createQueue = atomically (newTBQueue (fromIntegral $ configQueueCapacity config))
+    createSentryService
+      | configDisableSentryLogging config = pure Nothing
+      | otherwise = maybe (pure Nothing) (fmap Just . getCrashLogger) (configSentryDSN config)
```
fatho: I think this is just traverse on `Maybe`, i.e.

```haskell
  | otherwise = traverse getCrashLogger (configSentryDSN config)
```

server/src/HttpServer.hs
```diff
@@ -60,6 +64,8 @@ postModification :: (Scotty.ScottyError e, MonadIO m) => Core -> Store.Modificat
 postModification core op = do
   -- the parameter is parsed as type (), therefore only presence or absence is important
   durable <- maybeParam "durable"
+  crash <- maybeParam "crash"
+  if crash == Just () then Scotty.liftAndCatchIO (error "Error ") else pure ()
```
fatho: This looks like a leftover from debugging, I don't think we need or want this in production.

PR #56, opened 2019-10-25
Add a SQLite backend to Icepeak

rkrzr:
## Overview

This PR adds a new experimental SQLite storage backend to Icepeak, which can be used instead of the default file-based backend. The type of backend must now be specified as a CLI flag when starting up the server using either `--file` or `--sqlite`. Example:

`icepeak --sqlite --data-file icepeak.db --sync-interval 1s --journaling`

I made sure that all the unit tests and integration tests are still passing for both backends and I also upgraded the integration tests from Python 2.7 to Python 3.7 by swapping out the older `websocket` library with the newer `websockets` library (note, the extra 's' in the end).

## Implementation

The current implementation is *very* simple. The whole JSON value is written to a single BLOB field in the database and the whole value will be overwritten whenever `sync` is called (depending on the given `--sync-interval`). It is basically identical to what the file-based backend does. The only minor difference is that we don't need to write to a second file first and then do an atomic rename, since Sqlite already gives us transactional guarantees. There is also no change in how journaling works, since the current mechanism works just as well for the Sqlite backend.

There is currently no advantage of using this backend yet - but the point of this PR was also more to lay the groundwork for implementing a more interesting Sqlite backend in the future, e.g. one that can split a JSON object over multiple rows (one per key) or even over multiple tables, which would allow us to do partial writes, instead of having to write all data on every sync.

## Limitations

I have done no serious performance testing of this backend yet. Running `./stress_test.hs "1000"` (with the server configured to persisting data after every modification) currently fails consistently after approximately 300 requests with a 503. We should still look into why that is.


server/integration-tests/data-survives-crash.sh
```diff
@@ -7,11 +7,17 @@ fi
...
-echo '{}' > "$DATA_FILE"
+
+BACKEND_FLAGS=("--file" "--sqlite")
 
 ########## 1st Pass
-stack exec -- icepeak --data-file="$DATA_FILE" --sync-interval 60s --journaling > /dev/null &
+
+for item in ${BACKEND_FLAGS[*]}
+do
+echo ">>>>>>>> Testing backend: $item"
```
ruuda: This looks a lot like a Git conflict marker, it would be nice to use something else, to avoid confusion and also to search for the conflict marker when there is a conflict without hitting this false positive.

server/integration-tests/multiple_clients_test.py
```diff
@@ -1,48 +1,54 @@
...
-        result = connection.recv()
+        # print('receiving {}'.format(connection))
+        result = await connection.recv()
         parsed_result = json.loads(result)
 
         assert parsed_result == expected_result, 'Input data: {} is different from output data: {}'.format(
             expected_result, parsed_result)
 
-    print 'All 10 clients received {} successfully!'.format(expected_result)
+    print('All 10 clients received {} successfully!'.format(expected_result))
```
ruuda: We can use f-strings in Python 3.7.

server/integration-tests/multiple_clients_test.py
```diff
@@ -1,48 +1,54 @@
...
 
 
-def main():
+async def main():
```
ruuda: Oooh, fancy!

server/integration-tests/read_after_write_test.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+read_after_write_check('', json1, json1)
+read_after_write_check('', json2, json2)
+read_after_write_check('', json3, json3)
+read_after_write_check('/so', {'cool': {}}, json3)
+read_after_write_check('/so/hot', json1, json4)
+
+print('All read-after-write tests passed successfully!')
+
```
ruuda: There are trailing blank lines here.

server/integration-tests/sqlite_test.hs
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env stack
+-- stack --stack-yaml ../../client-haskell/stack.yaml runghc --package sqlite-simple --package aeson
+
+{-# OPTIONS_GHC -Wall #-}
+
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+{- | Test whether SQLite was installed with JSON support enabled by calling the "json" SQL
```
ruuda: This comment looks misplaced, was it supposed to go elsewhere?

server/integration-tests/using-different-ports.sh
```diff
@@ -1,4 +1,4 @@
-#/bin/bash
+#!/bin/bash
```
ruuda: Nice fix!

server/src/Config.hs
```diff
@@ -17,15 +18,15 @@ import qualified Data.List as List
...
 -- command-line arguments
 data Config = Config
   { configDataFile :: FilePath
   , configPort :: Int
     -- | Enables the use of JWT for authorization in JWT.
-  , configEnableJwtAuth :: Bool
-    -- | The secret used for verifying the JWT signatures. If no secret is
-    -- specified even though JWT authorization is enabled, tokens will still be
-    -- used, but not be verified.
+  , configEnableJwtAuth :: Bool -- | The secret used for verifying the JWT signatures. If no secret is -- specified even though JWT authorization is enabled, tokens will still be -- used, but not be verified.
```
ruuda: This line wrapping looks unintended.

server/src/Config.hs
```diff
@@ -111,6 +115,16 @@ configInfo environment = info parser description
     description = fullDesc <>
       header "Icepeak - Fast Json document store with push notification support."
 
+-- * Parsers
+
+storageBackend :: Parser StorageBackend
+storageBackend = fileBackend <|> sqliteBackend
```
ruuda: For backwards compatibility, it may be nice to default to `--file`.

server/src/Persistence.hs
```diff
@@ -7,14 +8,17 @@ module Persistence
   , PersistenceConfig (..)
   , getValue
   , apply
-  , load
-  , sync
+--  , load
+  , loadFromBackend
+--  , sync
```
ruuda: It looks like these comments can be removed now.

server/src/Persistence.hs
```diff
@@ -23,6 +27,7 @@ import           Data.Foldable
 import           Data.Text                  (Text)
 import qualified Data.Text                  as Text
 import           Data.Traversable
+import           Database.SQLite.Simple
```
ruuda: It would be nice to use a qualified import; for function like `execute`, `open`, and `field`, it is not immediately obvious that they relate to SQLite at the call site.

server/src/Persistence.hs
```diff
@@ -69,9 +76,94 @@ apply op val = do
...
+  fromRow = JsonRow <$> field
+
+loadSqliteFile :: PersistenceConfig -> IO (Either String PersistentValue)
+loadSqliteFile config = runExceptT $ do
+  liftIO $ forM_ (pcMetrics config) $ \m -> do
+    size <- getFileSize (pcDataFile config)
+    Metrics.setDataSize size m
+
+  value <- readSqliteData (pcDataFile config) (pcLogger config)
+  valueVar <- lift $ newTVarIO value
```
ruuda: This setup is duplicated here from `loadFile`, can we extract it so it is in one place?

server/src/Persistence.hs
```diff
@@ -69,9 +76,94 @@ apply op val = do
...
+  return val
+
+-- | Read and decode the Sqlite data file from disk
+readSqliteData :: FilePath -> Logger -> ExceptT String IO Store.Value
+readSqliteData filePath _logger = ExceptT $ do
+  -- read the data from SQLite
+  conn <- liftIO $ open filePath
+  -- TODO: We might want to have a primary key here after all. Then we could have multiple
+  -- JSON values next to each other.
+  liftIO $ execute_ conn "CREATE TABLE IF NOT EXISTS test (value BLOB)"
```
ruuda: It's a bit unexpected that read creates the table. I would expect reads to work against a read-only database, for example. Maybe we should move this into `loadSqliteFile` instead?

server/src/Persistence.hs
```diff
@@ -69,9 +76,94 @@ apply op val = do
...
+  jsonRows <- liftIO $ (query_ conn "SELECT * from test" :: IO [JsonRow])
+
+  case jsonRows of
+    -- if there is no data yet, we simply return the empty object. We do the same thing for the
+    -- file backend.
+    [] -> do
+      liftIO $ execute conn "INSERT INTO test (value) VALUES (?)" (Only $ Aeson.encode Aeson.emptyObject)
```
ruuda: Also this insert, why do we need it at all? If we don't write the value, we will default to `{}` next time anyway.

If we want to do a dummy insert at startup to ensure that we have write access, I think it is better to do that explicitly during startup. That way we can also deduplicate it between SQLite and the file backend.

server/src/Persistence.hs
```diff
@@ -69,9 +76,94 @@ apply op val = do
...
+
+    conn <- open filePath
+    -- we can always UPDATE here, since we know that there will be at least one row, since
+    -- we issue an INSERT when we load in an empty database
+    liftIO $ executeNamed conn "UPDATE test SET value = :value" [":value" := Aeson.encode value]
+
+    -- the journal is idempotent, so there is no harm if icepeak crashes between
+    -- the previous and the next action
+    for_ (pvJournal val) $ \journalHandle -> do
+      hSeek journalHandle AbsoluteSeek 0
```
ruuda: Why is a seek needed here? Does `hSetFileSize` only truncate from the current position?

fatho: That's copied form code I wrote when I introduced journaling. `hSetFileSize` does not change the file cursor, which means that the first write that follows will fill up the file with `\NUL` bytes up to the cursor, and then proceed to write normally.
Therefore, we need to manually seek to the start of the file.
In hindsight, that fact could've been documented better, but it seems we were a bit in a hurry back then when we fixed it, because it was rapidly filling the disk in production :see_no_evil:

ruuda: Ah, that explanation makes sense, but it would be nice to have that in a comment here indeed.

server/integration-tests/data-survives-crash.sh
rkrzr: True, fixed.

server/integration-tests/sqlite_test.hs
rkrzr: I updated the comment now.

server/integration-tests/using-different-ports.sh
rkrzr: It was actually a handy code inspection in Intellij that pointed this out.

server/src/Config.hs
rkrzr: Oh, that was an accident. Fixed.

server/src/Config.hs
rkrzr: Good idea, fixed.

server/src/Persistence.hs
arianvp: (Hi; I still am subscribed to this repo apparently. hehe. friendly drive-by comment)

I think this will also maybe cause a race condition.  Should probably rwap the `SELECT * from test` and `INSERT INTO`  into a `BEGIN TRANSACTION` and `COMMIT` no?  otherwise if two people modify the empty database at the same time, you might end up wiping out the changes

fatho: > otherwise if two people modify the empty database at the same time, you might end up wiping out the changes

Not sure if this is something we need to guard against with the single row single column backend. After all, while icepeak is running it will periodically update the value anyway, without first reading it, and thus overwriting any changes made concurrently.
It's the same as with the other backend: if I modify `icepeak.json` while Icepeak is running, then I cannot expect those changes to be persisted.

If possible, we should make the sqlite connection exclusive so that no concurrent modifications can happen while Icepeak is running.

server/src/Persistence.hs
rkrzr: Thanks for pointing that out. I have deduplicated this code now and added your explanation.

server/src/Persistence.hs
rkrzr: Yep, done!

server/src/Persistence.hs
rkrzr: I have added a `setupStorageBackend` function now, that is called at program start and that ensures that we can access the file, and that validates that it has the right format. If there is an error, it prints a nice error message and exits the program.

server/src/Persistence.hs
rkrzr: I moved the INSERT to `setupStorageBackend` now.
I agree with Fabian that we assume that Icepeak is the only writer to the Sqlite file - it's the same as for the file backend.

rkrzr:
I have finally gotten around to addressing all issues here ;)

Please take another look.

server/integration-tests/data-survives-crash.sh
ruuda: This has not been fixed?

server/integration-tests/multiple_clients_test.py
```diff
@@ -1,48 +1,53 @@
...
-test_url ='http://localhost:3000/so/cool'
+test_url = 'http://localhost:3000/so/cool'
 
-def validate_data(connections, expected_result):
+
+async def validate_data(connections, expected_result):
     for connection in connections:
-        print 'receiving {}'.format(connection)
-        result = connection.recv()
+        # print('receiving {}'.format(connection))
```
ruuda: Can this be removed?

server/integration-tests/sqlite_test.hs
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env stack
```
ruuda: What is this file supposed to test? I think we can assume that SQLite works. Is this a leftover experiment?

server/src/Config.hs
```diff
@@ -54,6 +59,7 @@ type EnvironmentConfig = [(String, String)]
 
 configParser :: EnvironmentConfig -> Parser Config
 configParser environment = Config
+  -- TODO: We should use a different default ("icepeak.db") when SQLite is chosen as the storage backend
```
ruuda: We might handle this elsewhere then, and not have a default in the parser? The option would be optional then.

server/src/Config.hs
ruuda: I don’t see a change here?

server/src/Persistence.hs
```diff
@@ -88,9 +145,61 @@ load config = runExceptT $ do
...
+-- | Read and decode the Sqlite data file from disk
+readSqliteData :: FilePath -> ExceptT String IO Store.Value
+readSqliteData filePath = ExceptT $ do
+  -- read the data from SQLite
+  conn <- liftIO $ open filePath
+  jsonRows <- liftIO $ (query_ conn "SELECT * from icepeak" :: IO [JsonRow])
+
+  case jsonRows of
+    -- if there is no data yet, we simply return the empty object. We do the same thing for the
+    -- file backend.
```
ruuda: This is no longer true, the empty object is now created elsewhere and read assumes it exists, right?

server/integration-tests/data-survives-crash.sh
rkrzr: I had fixed it in `data-survives-restart.sh`, but then forgotten to fix it here. Now it's fixed.

server/integration-tests/sqlite_test.hs
rkrzr: This was indeed an experiment. It was actually pretty useful to try a few different approaches on how to store the value in Sqlite. I could move it somewhere else, but I think it doesn't really hurt to leave it here.

server/src/Config.hs
rkrzr: I had forgotten to push it (was on a different machine).

server/src/Persistence.hs
rkrzr: Correct, I updated the comment.

PR #60, opened 2020-01-17
Build/package with Nix

duijf:
Add a Nix package definition for Icepeak.

```
$ cd server
$ nix build
```

### Background

Nix is a purely functional build system. It can be used to build software and their dependencies. Read [The Nix Ecosystem](https://nixos.wiki/wiki/Nix_Ecosystem) for a high level introduction.

This PR uses Nix to build Icepeak. Nix and the extended Nix ecosystem enable us to define the exact dependencies that were used in order to build a package.

Internally, Channable is starting to use Nix for more things. This PR is part of our efforts to gain more experience in using Nix for all types of software. This PR might also help us easily build and distribute icepeak binaries.

Read https://nixos.org/nixos/nix-pills/ for an introduction to the Nix language and to derivations.

### Background: What's in Nixpkgs?

Nixpkgs contains a large collection of software ranging from applications to development tooling. It also contains a set of Haskell packages (similar to Stackage).

The things you need to know are:

 - Nixpkgs is generally pinned to a specific commit for reproducibility. By convention, the entirety of nixpkgs is generally imported as `pkgs`.
 - Nixpkgs contains infrastructure for building Haskell packages. This functionality is available under `pkgs.haskellPackages`. This [attribute set](https://nixos.wiki/wiki/Nix_Expression_Language#Types) contains packages themselves, as well as functions for constructing more of them. The most notable one we use in this PR is `pkgs.haskellPackages.callPackage` and `pkgs.haskellPackages.mkDerivation`.

### Approach

This PR adds 4 new files:

 - Pin a specific version of Nixpkgs in `nix/nixpkgs.nix`. This helps with reproducibility.
 - Specify packages in `server.nix` and `nix/websockets.nix`. These files contain a function with type `AttrSet -> Derivation`. They expect a set of dependencies to be passed to them so they can eventually build a derivation out of them.
 - Wire everything up in `default.nix`. This is the trickiest bit.

### On `default.nix`

The short version: we only want to compile against a single version of a library like `websockets`. Therefore we need to override it on the Haskell package set from Nixpkgs. We cannot just pass the new version to `icepeak` itself as we might also transitively depend on `websockets` through something else.

The boilerplate in `default.nix` overrides `pkgs.haskellPackages.websockets` for `icepeak` **and** for all other packages in `pkgs.haskellPackages`. Nixpkgs has utilities which ensure that everything works out.

While we're at it, we also include `icepeak` in the overriden set of Haskell packages. That would only be useful if other packages from Nixpkgs ever add a dependency on Icepeak, but that's apparently good practice.

For more background and tutorials, I can recommend https://github.com/Gabriel439/haskell-nix/

### Results

We get a dynamic executable out of the build process which does what we want.

Next steps include:

 - Improving the `src` directive (currently this leads to excessive rebuilds)
 - Building this on CI
 - Upstreaming to Nixpkgs if we like.

### DIFF
--- a/server/default.nix
+++ b/server/default.nix
@@ -1,5 +1,17 @@
 let
-  pkgs = import ../nix/nixpkgs.nix {};
-  icepeak-server = pkgs.haskellPackages.callPackage ./server.nix {};
+  config = {
+    packageOverrides = pkgs: rec {
+      haskellPackages = pkgs.haskellPackages.override {
+        overrides = haskellPackagesNew: haskellPackagesOld: rec {
+          icepeak = haskellPackagesNew.callPackage ./server.nix { };
+          websockets = haskellPackagesNew.callPackage ../nix/websockets.nix { };
+        };
+      };
+    };
+  };
+
+  pkgs = import ../nix/nixpkgs.nix { inherit config; };
 in
-  icepeak-server
+  {
+    icepeak-server = pkgs.haskellPackages.icepeak;
+  }
--- a/server/default.nix
+++ b/server/default.nix
@@ -1,0 +1,5 @@
+let
+  pkgs = import ../nix/nixpkgs.nix {};
+  icepeak-server = pkgs.haskellPackages.callPackage ./server.nix {};
+in
+  icepeak-server
--- a/server/default.nix
+++ b/server/default.nix
@@ -1,3 +1,10 @@
+{ # We expect getPkgs to be a lambda that gives us a package set. Example:
+  # the lambda in `nixpkgs/default.nix` without any arguments passed to it.
+  # That allows us to pass in config later, while allowing users to pass in
+  # their own version of Nixpkgs.
+  getPkgs ? import ../nix/nixpkgs.nix
+}:
+
 let
   config = {
     packageOverrides = pkgs: rec {
@@ -10,7 +17,7 @@ let
     };
   };
 
-  pkgs = import ../nix/nixpkgs.nix { inherit config; };
+  pkgs = getPkgs { inherit config; };
 in
   {
     icepeak-server = pkgs.haskellPackages.icepeak;
--- a/server/server.nix
+++ b/server/server.nix
@@ -41,7 +41,7 @@
 
 mkDerivation {
   pname = "icepeak";
-  version = "0.6.2";
+  version = "0.6.3";
 
   # Opt in to hpack. We don't commit the cabal file in our repo currently.
   buildTools = [ hpack ];
--- a/server/server.nix
+++ b/server/server.nix
@@ -5,6 +5,7 @@
 , containers
 , directory
 , hashable
+, hpack
 , hspec
 , hspec-wai
 , http-types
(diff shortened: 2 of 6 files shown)


server/server.nix
```diff
@@ -0,0 +1,92 @@
...
+, unix
+, unordered-containers
+, uuid
+, wai
+, wai-extra
+, wai-middleware-prometheus
+, wai-websockets
+, warp
+, websockets
+}:
```
NunoAlexandre: What's the reason to have this list here not assigned to anything and then have the same packages listed again in `executableHaskellDepends`?

server/default.nix
```diff
@@ -0,0 +1,17 @@
...
+      haskellPackages = pkgs.haskellPackages.override {
+        overrides = haskellPackagesNew: haskellPackagesOld: rec {
+          icepeak = haskellPackagesNew.callPackage ./server.nix { };
+          websockets = haskellPackagesNew.callPackage ../nix/websockets.nix { };
+        };
+      };
+    };
+  };
+
+  pkgs = import ../nix/nixpkgs.nix { inherit config; };
```
rightfold: Would be nice if this were a parameter (the import can be its default) value so that you can use Icepeak from another project with a different nixpkgs if desires.

server/server.nix
duijf: Hey Nuno! If you're entirely new to Nix, I probably can't explain the entire thing from scratch here, but I'll try.

The quick summary is: this file defines a function which accepts an attribute set where we expect certain keys to be present.

```nix
# This is a function which accepts an attrset with keys 
# `foo` and `bar` and returns the `foo` key.
{ foo
, bar
}:
	foo
```

When we pass `./server.nix` to `callPackage` in `default.nix`, the `callPackage` function inspects the list of required attributes and wires these up with what it has in scope from Nixpkgs. This is a convention from Nixpkgs to express "I want this function to get the following arguments".

As you can see not every argument from the attrset goes into the same place. Some of the arguments go to `executableHaskellDepends`. Others go to `testHaskellDepends` or are used directly (`mkDerivation` or `stdenv.lib`)

Intro to the Nix language: https://nixos.org/nix/manual/#ch-expression-language

Packaging Haskell with Nix in https://github.com/Gabriel439/haskell-nix/

NunoAlexandre: Thank you so much, your explanation is clear and I get it now. Thanks also for the links, Nix can wait no more!

server/default.nix
duijf: I attempted something in 8c94826aa24b7616cf8bf51a441d8d9e19193bbf, but I'm not sure that's the best way to do this.

Is there a way to pass in new config / package overrides on an already imported set of nixpkgs?

duijf:
In the last two commits I added source filtering so we only really make the files that are required to build icepeak available to Nix (when in development). 

That also required me to add a call to hpack, as we do not track the cabal file in git currently.

server/default.nix
rightfold: Not sure. There is no overrideAttrs on the object returned by Nixpkgs.

server/server.nix
```diff
@@ -41,7 +41,56 @@
 mkDerivation {
   pname = "icepeak";
   version = "0.6.2";
```
rkrzr: Drive-by: Should be `0.6.3`

duijf: Ah, then the version number in `server/package.yaml` is out of date as well.

rkrzr: True, that needs to be updated as well then.

PR #61, opened 2020-01-30
Add nix derivation for icepeak Haskell client

fatho:
This adds a nix derivation for the icepeak-client Haskell library, analogously to the server.

Also, this builds the nix derivations on CI to make sure that they actually work.


.travis.yml
```diff
@@ -33,10 +33,17 @@ install:
...
 
 script:
   # Build and test.
   - (cd server && stack build -j2 --no-terminal)
   - (cd server && stack test  -j2 --no-terminal)
   - (cd client-haskell && stack build -j2 --no-terminal)
   - (cd client-haskell && stack test  -j2 --no-terminal)
+  # Check that the nix builds work
+  - nix-build --no-out-link -E '(import ./nix/nixpkgs.nix {}).haskellPackages.callPackage ./client-haskell/client.nix {}'
+  - nix-build --no-out-link ./server/default.nix
```
duijf: We could also create a `default.nix` in the repository root and use that to build both the client and the server

fatho: I would reserve a top-level `default.nix` for proper integration in the nix-ecosystem, e.g. a nixpkg overlay providing the two packages.
But since it also has to override the version of the `websockets` library, this is best left for a separate PR.

PR #63, opened 2020-02-05
Update to latest Nix snapshot

rkrzr:
We no longer need to override the websockets package, since the latest
Nix snapshot already contains it, making our derivation simpler.


duijf:
LGTM!

PR #64, opened 2020-03-10
Upgrade stack from LTS 14.23 to 15.3

rkrzr:
This also upgrades GHC from 8.6 to 8.8. There have been no breaking
changes and the tests are still passing. I did have to add sqlite-simple
and direct-sqlite as extra-deps, since they are no longer in the
stackage snapshot.

Why upgrade?

I have been on a [quixotic quest](https://github.com/NixOS/nixpkgs/pull/79155) to get Icepeak added to the Nixpkgs archive. And nixpkgs recently upgraded their supported GHC version from 8.6 to 8.8, and it's therefore necessary to upgrade here as well, because I can then release a new version of Icepeak on Hackage, which can then be downloaded by an automatic script `hackage2nix`, which will generate a Nix derivation for Icepeak, that is then added to Nixpkgs. :see_no_evil:

### DIFF
--- a/client-haskell/stack.yaml
+++ b/client-haskell/stack.yaml
@@ -1,1 +1,1 @@
-resolver: lts-14.23
+resolver: lts-15.3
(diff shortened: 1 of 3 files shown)


rkrzr:
Yes, that sounds like a good idea. I still have to figure out how to do that, so for now I'll go ahead and merge this and then we can still do that in a future PR.

PR #65, opened 2020-03-30
Add metrics with recommended Prometheus naming scheme

### LINKED ISSUE #23: Rename metrics to use recommended Prometheus naming scheme
The naming scheme is outlined [here](https://prometheus.io/docs/practices/naming/).

WJWH:
Pull request to fix #23. This PR:

- Expands the `IcepeakMetrics` data structure with two extra fields that follow the recommended Prometheus naming scheme.
- Changes the  `incrementDataWritten` and `setDataSize` to write to both the old and new metrics.
- Adds TODO comments indicating which lines can be removed later. The resulting lines are a bit long for my liking but that is all the more incentive to update the dashboard soon so that they can be removed :)

By adding the new metrics first, it'll become possible to update any dashboards (as mentioned in the linked issue) and then in a second PR the old metrics can be removed. I'm unsure what the best new name for the HTTP request counter metric would be. It could be something like `icepeak_http_requests_total`?

### DIFF
--- a/server/src/Metrics.hs
+++ b/server/src/Metrics.hs
@@ -25,21 +25,25 @@ countHttpRequest method status httpRequestCounter = withLabel httpRequestCounter
 
 
 data IcepeakMetrics = IcepeakMetrics
-  { icepeakMetricsRequestCounter  :: HttpRequestCounter
-  , icepeakMetricsDataSize        :: Gauge
-  , icepeakMetricsJournalSize     :: Gauge
-  , icepeakMetricsDataWritten     :: Counter
-  , icepeakMetricsJournalWritten  :: Counter
-  , icepeakMetricsSubscriberCount :: Gauge
+  { icepeakMetricsRequestCounter    :: HttpRequestCounter
+  , icepeakMetricsDataSize          :: Gauge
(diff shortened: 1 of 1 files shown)


server/src/Metrics.hs
```diff
@@ -25,21 +25,25 @@ countHttpRequest method status httpRequestCounter = withLabel httpRequestCounter
...
 
 data IcepeakMetrics = IcepeakMetrics
-  { icepeakMetricsRequestCounter  :: HttpRequestCounter
-  , icepeakMetricsDataSize        :: Gauge
-  , icepeakMetricsJournalSize     :: Gauge
-  , icepeakMetricsDataWritten     :: Counter
-  , icepeakMetricsJournalWritten  :: Counter
-  , icepeakMetricsSubscriberCount :: Gauge
+  { icepeakMetricsRequestCounter    :: HttpRequestCounter
+  , icepeakMetricsDataSize          :: Gauge -- TODO: can be removed after dashboard has been updated to use icepeak_data_size_bytes
```
rkrzr, line 29: You can put the TODOs above each line to avoid the long lines.

WJWH: I moved the comments to their own lines.

PR #66, opened 2020-04-08
Travis speed

WJWH:
I noticed when working on #65 that the travis CI system took quite long (16 minutes for my branch, between ~8 and ~30 minutes [in the build history on Travis](https://travis-ci.org/github/channable/icepeak/builds)). The Coronavirus also struck Almere, so I'm bored and decided to speed it up.

This PR:
- Splits the four tasks in the Travis CI file into their own job so that they can run in parallel. Travis allows parallel jobs even for open source projects.
- Optimizes the caching and setup for each of the resulting jobs. (Testing the server does not need the stack cache for the client and vice versa, for example)
- Changes the Travis base image from `c` to `minimal` ([link to docs](https://docs.travis-ci.com/user/languages/minimal-and-generic/)). This does not really matter too much for this project but is a little bit cleaner. It also meshes better with the pre-existing comment.

As a result, the wall clock time is now closer to five minutes. Be warned btw, every named job on Travis gets its own cache, which persists between builds. This means that the first time after this gets merged, performance will be horrible as everything is uncached. It'll get better afterwards though, see the attached screenshot.
Screenshot from 2020-04-08 22-10-30
It could be sped up by a few more minutes if Travis had images with stack preinstalled, the cache contains so many files that even if nothing changed it takes >80 seconds to figure that out. Downloading the entire cache is not fast either.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,29 +26,43 @@ before_install:
 - export PATH=$HOME/.local/bin:$PATH
 - travis_retry curl -L https://www.stackage.org/stack/linux-x86_64 | tar xz --wildcards --strip-components=1 -C ~/.local/bin '*/stack'
 
-install:
-  # Install the compiler. Pass --no-terminal to tell Stack it is not running in
-  # an interactive terminal, so it prints messages sequentially instead of
-  # updating them. The latter is less spammy but Travis does not support these
-  # interactive terminal capabilities.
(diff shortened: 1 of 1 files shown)


WJWH:
Btw, cool blog post about the compact regions this morning! :slightly_smiling_face:

rightfold:
Hmm, I restarted the build and after it finished, GitHub is still saying that it is pending.

PR #67, opened 2020-12-03
Update nixpkgs snapshot

rkrzr:
This upgrades GHC from 8.8.2 to 8.8.4. This is necessary to stay compatible with the
master branch of nixpkgs. Icepeak was marked as broken there, since it
couldn't be built any more with the GHC and libraries available there.

Also update the Stack LTS release from 15.3 to 16.23.

This will then allows us to cut a 0.7.4 release, and get that included into nixpks to unbreak it.

### DIFF
--- a/client-haskell/src/Icepeak/Client.hs
+++ b/client-haskell/src/Icepeak/Client.hs
@@ -43,6 +43,7 @@ import Control.Retry (RetryPolicyM, recovering)
 import Data.Aeson (ToJSON)
 import Data.ByteString (ByteString)
 import Data.Foldable (toList)
+import Data.Semigroup ((<>))
 import Data.Text (Text)
 import Data.Word (Word16)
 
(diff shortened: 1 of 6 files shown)


client-haskell/src/Icepeak/Client.hs
```diff
@@ -43,7 +43,6 @@ import Control.Retry (RetryPolicyM, recovering)
 import Data.Aeson (ToJSON)
 import Data.ByteString (ByteString)
 import Data.Foldable (toList)
-import Data.Semigroup ((<>))
```
fatho, line 46: This means that we're probably no longer compatible with old `base` packages where `(<>)` was not exported by the prelude. This either needs conditional compilation, or the version bound needs to be tightened. Or we accept and silence the warning.

fatho (commented):
One small note about backwards compatibility:

client-haskell/src/Icepeak/Client.hs
rkrzr: Good point. We can just accept the warning.

PR #68, opened 2021-03-27
Upgrade stack from LTS 16.23 to 17.2 (GHC 8.8 -> 8.10)

rkrzr:
This requires bumping the nixpkgs revision. The latest nixpkgs-unstable revision contains GHC 8.10.3, which we are now using. There were no breaking changes in any of our dependencies.


NunoAlexandre (approved):
:climbing:

fatho (approved):

PR #69, opened 2021-08-05
Ci/semaphore setup

Riscky:
This PR removes the Travis CI configuration in favor of Semaphore CI.
This checks one of the steps of channable/devops#3788


.semaphore/semaphore.yml
```diff
@@ -0,0 +1,88 @@
...
+          # Download and verify Nix installation script.
+          - curl -o install-nix-2.3.15 https://releases.nixos.org/nix/nix-2.3.15/install
+          - sha256sum --check .semaphore/install-nix-2.3.15.sha256
+
+          # Restore `/nix` cache. Create the directory first, otherwise we encounter
+          # permission errors. We do this because the Semaphore cache is faster than
+          # both Cachix and cache.nixos.org.
+          - "sudo mkdir /nix"
+          - "sudo chown -R semaphore:semaphore /nix"
+          - "cache restore nix-store-"
```
fatho, line 65: By restoring *any* nix-store cache we have, the nix store will only ever grow until we run out of disk at some point. Since we're not running a lot of builds on this repo, this is probably fine, but something to keep in mind. In our other nix-based projects we only restore the nix-store cache from the same day, so the first build on each day takes a bit longer.

.semaphore/semaphore.yml
```diff
@@ -0,0 +1,88 @@
...
+            - "(cd server && stack build -j2 --no-terminal)"
+            - "(cd server && stack test  -j2 --no-terminal)"
+
+      epilogue:
+        always:
+          commands:
+            # Store a copy of the stack installation folder
+            - "cache store stack-$(date -u -Idate) ~/.stack"
+
+  - name: "Build"
```
fatho, line 50: It seems that in semaphore, there's a dependency from this step to the previous one, but those are actually independent, so we could speed up the build a bit more.

This second step ensures that the nix build keeps working (which is how we use icepeak internally), while the other step makes sure that the pure stack build keeps working (which is still a lot more common in the Haskell world).

fatho (commented):
Very nice, just some remarks:

.semaphore/semaphore.yml
Riscky: I'm unsure why Semaphore thinks that there is a dependency, I'll see what I can do

.semaphore/semaphore.yml
Riscky: You have a point, but for this repo I'll leave it as is

.semaphore/semaphore.yml
Riscky: Fixed by making explicit that there are no dependencies :)

.semaphore/semaphore.yml
```diff
@@ -0,0 +1,92 @@
+---
+version: "v1.0"
+name: "Vaultenv CI Pipeline"
```
bertptrs, line 3: Nit: the CI pipeline is called `Vaultenv CI Pipeline`. I can guess why this is the cause, but we probably do want to change it.

rkrzr (approved):

.semaphore/semaphore.yml
Riscky: oops

Issue #71, opened 2021-08-13
Vulnix report: verify that libxml is not a runtime dependency


Riscky:
There are a few known vulnerabilities in the version of libxml2 that we depend on.
Make sure these won't hit us, either by checking that the library is not called on runtime, or by updating the library to a newer  version.

Issue #70, opened 2021-08-13
Vulnix report: update raven-haskell


Riscky:
Raven-haskell depends on a lot of outdated libraries with known vulnerabilities.
There has not been an update of raven-haskell since February 2020, so this might require forking the library.

PR #72, opened 2021-08-13
Add Vulnix as a nightly CI job

Riscky:
This PR is related to channable/devops/issues/3773

This PR adds a nightly run CI job that checks if there are no known CVE's in our dependencies, using Vulnix.
Vulnix is setup up to use a large ignorelist that makes sure that the CI does not start failing immediately.
There are issues created for the package updates we have to perform: #70 and #71, which are linked in the ignorelist to the right CVE's

I have tested the CI setup in a seperate branch. You can find the result here: https://channable.semaphoreci.com/workflows/d7ded8d6-be3e-4057-8054-1e05ab34a2b9?pipeline_id=333c929f-62b8-4d19-ae8b-9a0c22d52a6c

### DIFF
--- a/.semaphore/vulnix-nightly.yml
+++ b/.semaphore/vulnix-nightly.yml
@@ -1,0 +1,65 @@
+---
+version: "v1.0"
+name: "Icepeak CI Pipeline"
+
+agent:
+  machine:
+    type: "e1-standard-2"
+    os_image: "ubuntu2004"
+
+# global_job_config:
+  # TODO: create HC item
+  # env_vars:
+  #   - name: "CRONJOB_HEALTHCHECK_URL"
+  #     value: "https://hc-ping.com/39f74a9c-4dad-46ca-a200-d2cdee5b8533"
+
+blocks:
+  - name: "Run Vulnix"
+    dependencies: []
+
+    task:
+      prologue:
+        commands:
+          - "checkout"
+
+          # Download and verify Nix installation script.
+          - curl -o install-nix-2.3.15 https://releases.nixos.org/nix/nix-2.3.15/install
+          - sha256sum --check .semaphore/install-nix-2.3.15.sha256
+
+          # Restore `/nix` cache. Create the directory first, otherwise we encounter
+          # permission errors. We do this because the Semaphore cache is faster than
+          # both Cachix and cache.nixos.org.
+          # This cache should only contain Vulnix and its dependencies
+          - "sudo mkdir /nix"
+          - "sudo chown -R semaphore:semaphore /nix"
+          - "cache restore nix-store-"
+
+          # For some reason, Semaphore CI sets this variable, but it causes the nix installation to fail
+          - unset LD_LIBRARY_PATH
+
+          # Install Nix and source the shell configuration immediately.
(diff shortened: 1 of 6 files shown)


.semaphore/vulnix-nightly.yml
```diff
@@ -0,0 +1,65 @@
...
+
+agent:
+  machine:
+    type: "e1-standard-2"
+    os_image: "ubuntu2004"
+
+global_job_config:
+  env_vars:
+    - name: "CRONJOB_HEALTHCHECK_URL"
+      value: "https://hc-ping.com/d80327c2-4cbc-4986-b4bb-c43002938f00"
```
maartenberg, line 13: Because this repository is public I'd prefer to not have this healthcheck URL visible. Please delete and regenerate the healthcheck, then we can add it as a secret in Semaphore.

.semaphore/vulnix-nightly.yml
```diff
@@ -0,0 +1,65 @@
...
+            - ".semaphore/report-cronjob-start.sh"
+            - "nix-build --no-out-link ./server/default.nix"
+            - "(cd server && nix-shell --run 'vulnix -w vulnix-ignorelist.toml $(nix path-info)')"
+
+      epilogue:
+        always:
+          commands:
+            # Store a copy of the nix store. This will be refreshed daily, which
+            # is more than sufficient for this repo.
+            - "cache store nix-store-vulnix$(date -u -Idate) /nix"
```
maartenberg, line 55: Is there a reason to make this job use a different Nix store cache than the regular CI runs?

maartenberg (changes requested):

.semaphore/vulnix-nightly.yml
maartenberg: Also note that if this is required then `vulnix` should preferably be at the start of the cache key: if Semaphore can't find a cache entry with a given key it'll look for a cache entry matching the prefix until a dash, so if `nix-store-1234-56-78` doesn't exist it looks for any key matching `nix-store-1234-56-*`, then `nix-store-1234-*`, and so on.
This means that if there is no `nix-store-vulnixwhatever` entry it can also restore a regular `nix-store-whatever` entry, which might not be what you want.

Riscky: This is a relic from an experiment where I tried to run Vulnix without having to build the project. I'll update to the correct path.

server/vulnix-ignorelist.toml
```diff
@@ -0,0 +1,117 @@
...
+"""
+issue_url = "https://github.com/channable/icepeak/issues/70"
+
+["flex"]
+cve = ["CVE-2019-6293"]
+comment = "This CVE describes a bug. As far as we know this library (a parser) is not called at runtime, which should prevent any issues"
+
+["glibc"]
+cve = ["CVE-2021-27645", "CVE-2021-33574", "CVE-2021-35942"]
+comment = "glibc is not a runtime dependency"
```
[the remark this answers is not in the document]
Riscky: Oh you're right, I'm not sure what I missed here

PR #77, opened 2021-09-21
Reduce server derivation's closure size

maartenberg:
This PR adds a new derivation `server/binaries-only.nix`.
This file takes the regular server build as input and strips unneeded Nix dependencies from it by removing the derivation's `lib` folder as suggested by @fatho.
Doing this reduces the closure size reported by `nix-tree` from around 3 GiB to around 50 MiB.

By using the `binaries-only` derivation instead of the `default.nix` we should be able to reduce the disk space needed for Icepeak in CI and the size of the Icepeak package we use on our servers.

Q: I made this into a separate derivation because I wasn't sure if there's any use for the metadata this package removes. Is there ever any use for this? If not then I think we can apply the `cp` trick in `default.nix` directly.


fatho:
> Q: I made this into a separate derivation because I wasn't sure if there's any use for the metadata this package removes. Is there ever any use for this?

Seems that we already build it with `isLibrary = False`, but it still includes the metadata. 

https://github.com/channable/icepeak/blob/1fdcefb8adaebc9c9057e24273571a6e84dd964a/server/server.nix#L101-L102

I don't think there is any use for that (since you cannot use it as a haskell dependency in another project anyway). 
Therefore, we could simplify it even more and simply `rm` the build-info directory (not sure what the exact name is) in a `postInstall` step of the derivation itself, without even creating a new derivation.

fatho (approved):
LGTM

maartenberg:
@OpsBotPrime merge

PR #79, opened 2021-12-04
Upgrade stack from LTS 17.2 to 18.18

rkrzr:
This required bumping nixpkgs. No code changes were required.

We now include stack in the nix environment, so that we don't have
to depend on stack being installed externally.


maartenberg (approved):
LGTM!

rkrzr:
@OpsBotPrime merge

PR #81, opened 2022-08-16
Upgrade from LTS 18.18 to 19.17 and fix all breaking changes

rkrzr:
This PR upgrades the stackage snapshot from 18.18 to 19.17.

This implies a GHC upgrade from 8.10 to 9.0.

There were breaking changes in two libraries:

- `Web.JWT` has split the Signer datatype into two versions: EncodeSigner and VerifySigner. The former is used to encode
newly minted tokens, while the latter is used to verify signed tokens.

- `Data.Aeson` has hidden the `HashMap` type behind a `KeyMap` type which only exposes `alterF` instead of `alter` and `adjust`. I added the last two functions to keep the rest of the code the same.

All the tests are passing again.

### DIFF
--- a/server/app/IcepeakTokenGen/Main.hs
+++ b/server/app/IcepeakTokenGen/Main.hs
@@ -13,7 +13,7 @@ import           AccessControl
 import           JwtAuth
 
 data Config = Config
-  { configJwtSecret      :: Maybe JWT.Signer
+  { configJwtSecret      :: Maybe JWT.EncodeSigner
   , configExpiresSeconds :: Maybe Integer
   , configWhitelist      :: [AuthPath]
   }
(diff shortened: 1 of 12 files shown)


server/app/IcepeakTokenGen/Main.hs
```diff
@@ -13,7 +13,7 @@ import           AccessControl
 import           JwtAuth
 
 data Config = Config
-  { configJwtSecret      :: Maybe JWT.Signer
+  { configJwtSecret      :: Maybe JWT.EncodeSigner
```
alex-mckenna, line 16: Maybe it would be a good idea to include a comment for this field so it's clear what it means if an `EncodeSigner` is not provided (similar to the comment already in `server/src/Config.hs`

alex-mckenna (approved):
LGTM, nice that the needed fixes were few and pretty mechanical :tada:

Issue #85, opened 2022-12-01
Add additional metrics


fatho:
Icepeak could use a few more metrics, namely for
- [x] the current length of the internal command queue.
- [x] the duration of the `Sync` operation that writes the current state to disk and flushes the journal
- [x] the duration of HTTP request handling

PR #86, opened 2022-12-09
New metrics

DiegoDiverio:
PR associated with issue https://github.com/channable/icepeak/issues/85.

### DIFF
--- a/server/src/HttpServer.hs
+++ b/server/src/HttpServer.hs
@@ -2,7 +2,6 @@
 
 module HttpServer (new) where
 
-import Control.Applicative (liftA2)
 import Control.Concurrent.MVar (newEmptyMVar, takeMVar)
 import Control.Monad
 import Control.Monad.IO.Class
@@ -26,30 +25,17 @@ import qualified Store
 import qualified Core
 import qualified Metrics
 
-import qualified Prometheus
-
-newtype Action e m a = Action { runAction :: Scotty.ActionT e m a }
-
-instance (Scotty.ScottyError e, MonadIO m) => Functor (Action e m) where
-  fmap f = Action . fmap f . runAction
-instance (Scotty.ScottyError e, MonadIO m) => Applicative (Action e m) where
-  pure = Action . return
-  liftA2 f a b = Action $ liftA2 f (runAction a) (runAction b)
-instance (Scotty.ScottyError e, MonadIO m) => Monad (Action e m) where
-  m >>= f = Action $ runAction m >>= runAction . f
-instance (Scotty.ScottyError e, MonadIO m) => MonadIO (Action e m) where
-  liftIO = Action . Scotty.liftAndCatchIO
-instance (Scotty.ScottyError e, MonadIO m) => Prometheus.MonadMonitor (Action e m) where
-  doIO = liftIO
-
 new :: Core -> IO Application
 new core =
   scottyApp $ do
-    -- First we check whether the request HTTP method is a recognised HTTP method.
+    -- First middleware logs the time Icepeak takes to handle a request as a
+    -- Prometheus metric.
+    middleware $ measureRequestDuration core
+    -- Then we check whether the request HTTP method is a recognised HTTP method.
     -- Any arbitrary ByteString is accepted as a request method and we store those 
     -- in the exposed metrics, this is a DoS vector.
     middleware canonicalizeHTTPMethods
-    -- Second middleware is the metrics middleware in order to intercept
+    -- Third middleware is the metrics middleware in order to intercept
     -- all requests and their corresponding responses
     forM_ (coreMetrics core) $ middleware . metricsMiddleware
     -- Exit on unknown HTTP verb after the request has been stored in the metrics.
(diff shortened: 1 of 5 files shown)


server/src/HttpServer.hs
```diff
@@ -25,6 +26,22 @@ import qualified Store
...
+  fmap f = Action . fmap f . runAction
+instance (Scotty.ScottyError e, MonadIO m) => Applicative (Action e m) where
+  pure = Action . return
+  liftA2 f a b = Action $ liftA2 f (runAction a) (runAction b)
+instance (Scotty.ScottyError e, MonadIO m) => Monad (Action e m) where
+  m >>= f = Action $ runAction m >>= runAction . f
+instance (Scotty.ScottyError e, MonadIO m) => MonadIO (Action e m) where
+  liftIO = Action . Scotty.liftAndCatchIO
+instance (Scotty.ScottyError e, MonadIO m) => Prometheus.MonadMonitor (Action e m) where
+  doIO = liftIO
```
fatho, line 43: What's this needed for?

server/src/HttpServer.hs
```diff
@@ -49,22 +66,25 @@ new core =
     when (configEnableJwtAuth $ coreConfig core) $
       middleware $ jwtMiddleware $ configJwtSecret $ coreConfig core
 
-    get (regex "^") $ do
+    get (regex "^") $ measureRequestDuration core $ do
```
fatho, line 69: This works, but with WAI applications, there's generally a nicer way of doing this: middlewares.
With scotty, they're a bit more hidden than in a plain warp app, but you can already see one being defined above.

With that, you get access to the lower levels, below scotty's own abstractions like `ActionT` which might simplify all this a bit.

You can also look at warpmachine for an inspiration, but it's a bit more complex than what we need here: https://github.com/channable/warpmachine/blob/fa12cf21916619e14734ed656b37691a72ae7f7d/src/Warpmachine.hs#L72-L76

server/src/Metrics.hs
```diff
@@ -51,9 +54,17 @@ createAndRegisterIcepeakMetrics = IcepeakMetrics
...
+  <*> register (histogram (Info "icepeak_sync_duration" "Duration of a Sync command.")
+                          syncBuckets)
+  <*> register (histogram (Info "icepeak_http_req_handling_duration"
+                                "Duration of the handling of an HTTP request.")
+                          httpBuckets)
   where
     requestCounter = counter (Info "icepeak_http_requests"
                                    "Total number of HTTP requests since starting Icepeak.")
+    syncBuckets = exponentialBuckets 0.0000008 1.2 20
+    httpBuckets = exponentialBuckets 0.00001 1.5 20
```
fatho, line 67: These seem to small if I understand the docs of `exponentialBuckets` correctly. I think anything smaller than 1ms is probably not interesting for us. In other services, for HTTP request handler durations we seem to use [`defaultBuckets`](https://hackage.haskell.org/package/prometheus-client-1.1.0/docs/src/Prometheus.Metric.Histogram.html#defaultBuckets).

For syncing, we might need our own scale though, the range we care about is probably between 1ms and 1s.

fatho (changes requested):
Congrats on your first PR :tada: 

I left some suggestions below for simplifying things, but overall it looks already quite solid.

server/src/Metrics.hs
```diff
@@ -51,9 +54,17 @@ createAndRegisterIcepeakMetrics = IcepeakMetrics
                               "Total number of bytes written to the journal so far."))
   <*> register (gauge
     (Info "icepeak_subscriber_count" "Number of websocket subscriber connections."))
+  <*> register (gauge (Info "icepeak_internal_queue_length" "Length of the internal queue."))
```
ruuda, line 57: If you make a gauge that tells you the current queue length, that tells you nothing about throughput. If Prometheus samples twice and it finds 0 both times, it could be that just nothing happened in the meantime, or it could be that 1000 items got enqueued and then dequeued in time before the next scrape. If the queue length starts growing, is it because the producer is suddenly producing more, or because the consumer stopped/slowed down consuming?

If instead you add two counters, one for the number of items added to the queue, and one for the number of items removed from the queue, then you can compute the queue length as `items_added_total - items_removed_total`, but in addition, you can compute the throughput of the producer and consumer.

You do need to be very careful to record every add/remove though, if you miss one, then `added - removed` can go out of sync with the actual queue length.

server/src/HttpServer.hs
DiegoDiverio: In [observeDuration](https://hackage.haskell.org/package/prometheus-client-1.1.0/docs/Prometheus.html#v:observeDuration), we need the monad to be an instance of `MonadMonitor`, that's why I wrapped it. If I take the middleware way, I *may* not have to do this though. I'll check it out.

server/src/Metrics.hs
```diff
@@ -51,9 +55,19 @@ createAndRegisterIcepeakMetrics = IcepeakMetrics
...
     (Info "icepeak_subscriber_count" "Number of websocket subscriber connections."))
+  <*> register (counter (Info "icepeak_internal_queue_items_added"
+                              "Total number of items added to the queue."))
+  <*> register (counter (Info "icepeak_internal_queue_items_removed"
+                              "Total number of items removed from the queue.."))
+  <*> register (histogram (Info "icepeak_sync_duration" "Duration of a Sync command.")
+                          syncBuckets)
+  <*> register (histogram (Info "icepeak_http_req_handling_duration"
+                                "Duration of the handling of an HTTP request.")
+                          defaultBuckets)
```
fatho, line 66: It might be a nice idea to split those per method and status, like the requestCounter metric in here.

Since a histogram also counts, I think we can actually deprecate the request counter metric eventually.

server/src/HttpServer.hs
```diff
@@ -28,11 +28,14 @@ import qualified Metrics
 new :: Core -> IO Application
 new core =
   scottyApp $ do
-    -- First we check whether the request HTTP method is a recognised HTTP method.
+    -- First middleware logs the time Icepeak takes to handle a request as a
+    -- Prometheus metric.
+    middleware $ measureRequestDuration core
```
fatho, line 33: When doing the split into status and method, we need to move this below the http method sanitization.

fatho (commented):
I have one more idea looking at the already existing metrics:

server/src/Metrics.hs
```diff
@@ -61,16 +54,19 @@ createAndRegisterIcepeakMetrics = IcepeakMetrics
...
   <*> register (histogram (Info "icepeak_sync_duration" "Duration of a Sync command.")
                           syncBuckets)
-  <*> register (histogram (Info "icepeak_http_req_handling_duration"
-                                "Duration of the handling of an HTTP request.")
-                          defaultBuckets)
   where
-    requestCounter = counter (Info "icepeak_http_requests"
-                                   "Total number of HTTP requests since starting Icepeak.")
-    syncBuckets = exponentialBuckets 0.001 2 12
+    requestHistogram = histogram (Info "icepeak_http_requests"
```
fatho, line 58: We should choose a different name following the naming convention for prometheus metrics: https://prometheus.io/docs/practices/naming/

Among the names they give as an example, they give `http_request_duration_seconds` which I think is applicable here. Note that an application prefix in the name is explicitly not included since HTTP requests are a generic domain.
We can still distinguish them because prometheus will remember which service it scraped a metric from.

fatho (changes requested):
A very small request, but code-wise it looks good!

Issue #87, opened 2022-12-09
Add some libraries to Nix dependencies


DiegoDiverio:
When trying to build with stack, I ran into a linking problem due to the absence of GMP on my system. I solved the problem with installing GMP via my system's package manager, but it would be better to have it via Nix. There is also an issue with python dependencies (websockets, etc.) such as the ones used for integration testing.

PR #88, opened 2022-12-12
Add python dependencies in Nix configuration files for the server.

DiegoDiverio:
Related to issue https://github.com/channable/icepeak/issues/87. 
I added the python dependencies needed for the integration tests, but didn't add GMP to the list since Icepeak builds successfully using `stack build --nix`.

### DIFF
--- a/server/default.nix
+++ b/server/default.nix
@@ -1,1 +1,3 @@
+{}:
 (import ./icepeak-server.nix {}).icepeak-server-env
+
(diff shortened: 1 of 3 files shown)


server/default.nix
```diff
@@ -1,15 +1 @@
...
-}:
-
-let
-  pkgs = getPkgs { };
-  icepeak-server = pkgs.haskellPackages.callPackage ./server.nix { };
-in
-  {
-    icepeak-server = icepeak-server;
-    stack = pkgs.haskellPackages.stack;
-  }
```
fatho, line 15: We might be relying on this file in this form in other projects that pull in icepeak.

Would it be an idea to introduce the development environment as part of the top-level default.nix that we already have?

fatho (approved):

Issue #91, opened 2023-01-18
Log information for Sync operation


DiegoDiverio:
We should log some information when the Sync operation runs, first and foremost the time it takes to complete and the moment at which it started. Having this extra information will make debugging performance issues easier.

PR #92, opened 2023-01-18
Log the time it takes to sync the file on disk.

DiegoDiverio:
Fixes https://github.com/channable/icepeak/issues/91. Didn't add the timestamp because we can see it with other tools anyway.

### DIFF
--- a/server/src/Persistence.hs
+++ b/server/src/Persistence.hs
@@ -30,6 +30,7 @@ import qualified Data.Text                  as Text
 import           Data.Traversable
 import           Database.SQLite.Simple     (FromRow (..), NamedParam (..), Only (..), execute,
                                              execute_, executeNamed, field, open, query_)
+import qualified System.Clock               as Clock
 import           System.Directory           (getFileSize, renameFile)
 import           System.Exit                (die)
 import           System.IO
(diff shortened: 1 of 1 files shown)


server/src/Persistence.hs
```diff
@@ -155,7 +156,12 @@ loadFromBackend backend config = runExceptT $ do
   return val
 
 syncToBackend :: StorageBackend -> PersistentValue -> IO ()
-syncToBackend File pv = syncFile pv
+syncToBackend File pv = do
+    start <- Clock.getTime Clock.Monotonic
+    syncFile pv
+    end <- Clock.getTime Clock.Monotonic
+    let time = Clock.toNanoSecs $ (Clock.diffTimeSpec end start) `div` 1000000
```
fatho, line 163: In the end, it probably doesn't matter, but isn't the order of operations conceptually wrong here?

The `Clock.toNanoSecs` spans the whole division, so despite its name, it returns milliseconds.

Just removing the $ should result in the same output in a less confusing way:  ``Clock.toNanoSecs (Clock.diffTimeSpec end start) `div` 1000000``

fatho (commented):
Thanks, Just a small thing:

server/src/Persistence.hs
DiegoDiverio: Yep, it's conceptually better without the `$`. I'll remove that!

fatho (approved):
:shipit:

DiegoDiverio:
@OpsBotPrime merge

OpsBotPrime:
Rebased as 16aebfc9095da2239ec5912b85b0e3a1fc09a26d, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/2cd25e54-577c-48ee-90f2-7423218aab9f?pipeline_id=fbb879a3-fbce-4901-bfb9-78a7cb2c2d07) started.

PR #93, opened 2023-02-13
Enable log toggling for Sync operations.

DiegoDiverio:
Before releasing, so not to break past behaviour, I preferred to add toggling for the Sync operation log.


fatho (approved):
Good idea, LGTM!

PR #98, opened 2023-03-02
Basic fix for avoiding blocking of the consumer thread.

DiegoDiverio:
We want to avoid dependency of one thread to a blocking queue, by stopping writing in it if it's full, thus discarding the update. This is not the best model and needs to be improved on, but it can serve as a fix in some cases.


fatho (approved):
LGTM, but before cutting a new release, it'd be good to also add the metrics so that we at least have a way of observing that this happened.

PR #99, opened 2023-03-02
Add metrics on WebSocket queue

DiegoDiverio:
We wanted to observe how this queue grows and shrinks over time. The metrics are available on the Prometheus endpoint.

### DIFF
--- a/server/src/Core.hs
+++ b/server/src/Core.hs
@@ -177,9 +177,12 @@ postUpdate path core = do
     full <- isFullTBQueue (coreUpdates core)
     -- In order not to block the reader thread, and subsequently stop processing coreQueue,
     -- we don't send new updates to subscribers if coreUpdates is full.
-    when (not full) $ writeTBQueue (coreUpdates core) (Just $ Updated path value)
+    unless full $ writeTBQueue (coreUpdates core) (Just $ Updated path value)
     return full
-  when (not isWsQueueFull) $ for_ (coreMetrics core) Metrics.incrementWsQueueAdded
+  for_ (coreMetrics core) $
+    if isWsQueueFull
+    then Metrics.incrementWsSkippedUpdates
+    else Metrics.incrementWsQueueAdded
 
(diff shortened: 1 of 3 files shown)


server/src/Metrics.hs
```diff
@@ -31,6 +31,8 @@ data IcepeakMetrics = IcepeakMetrics
   , icepeakMetricsQueueAdded        :: Counter
   , icepeakMetricsQueueRemoved      :: Counter
   , icepeakMetricsSyncDuration      :: Histogram
+  , icepeakMetricsWsQueueAdded      :: Counter
+  , icepeakMetricsWsQueueRemoved    :: Counter
```
fatho, line 35: Shall we just add the counter metric for skipped updates in this PR, too?

server/src/Core.hs
```diff
@@ -169,14 +171,15 @@ runCommandLoop core = go
...
-    else writeTBQueue (coreUpdates core) (Just $ Updated path value)
+postUpdate path core = do
+  isWsQueueFull <- atomically $ do
+    value <- Persistence.getValue (coreCurrentValue core)
+    full <- isFullTBQueue (coreUpdates core)
+    -- In order not to block the reader thread, and subsequently stop processing coreQueue,
+    -- we don't send new updates to subscribers if coreUpdates is full.
+    when (not full) $ writeTBQueue (coreUpdates core) (Just $ Updated path value)
+    return full
+  when (not isWsQueueFull) $ for_ (coreMetrics core) Metrics.incrementWsQueueAdded
```
fatho, line 182: Then here you would just decide which counter to increment based on whether it's full or not.

fatho (commented):
Basically LGTM, but one suggestion:

server/src/Metrics.hs
DiegoDiverio: Good thinking, will add!

fatho (approved):
LGTM

PR #100, opened 2023-03-23
Import haskell-language-server in `default.nix` and update GHC version to 9.2.7

DiegoDiverio:
Contents of this PR:
- Update GHC to version 9.2.7. This choice of version is motivated by the fact that it is the last version with a `stack` LTS resolver.
- Niv pin update.
- Import of `haskell-language-server` for the server.
- Update of `stack` resolver to `lts-20.15`.

I didn't add `enable: true` to the `nix` section of `stack.yaml` for non-Nix developers to be able to use it seamlessly, but it is really handy (maybe even necessary for some cases) when using a Nix environment (and `haskell-language-server`).

PR #101, opened 2023-03-23
Change threading model for subscribers

DiegoDiverio:
The previous model for subscribers was monolithic: one thread was managing all the subscribers, and if one of them failed to answer, it could block the entire subscribers mechanism. We want to avoid this "single point of failure" kind of architecture, so we opted for a different thread model.

Now, for *each* subscriber, we have a dedicated thread that handles sending the updates to them. This thread is killed on timeout or unsubscribing. Icepeak's main thread feeds these threads through dedicated bounded queues (one per subscriber).

This PR starts as a draft because I want https://github.com/channable/icepeak/pull/100 merged first.

### DIFF
--- a/server/src/Config.hs
+++ b/server/src/Config.hs
@@ -45,6 +45,8 @@ data Config = Config
   , configStorageBackend :: StorageBackend
   -- | Enable logging of Sync duration
   , configSyncLogging :: Bool
+  -- | Delay between two pings to the client (useful in order to keep the connexion alive)
+  , configWebSocketPingInterval:: Int
   }
 
 data MetricsConfig = MetricsConfig
@@ -105,6 +107,11 @@ configParser environment = Config
   <*> storageBackend
   <*> switch (long "sync-log" <>
              help "Enable logging the duration of Sync operations.")
+  <*> option auto
+       (long "websocket-ping-interval" <>
+        metavar "WS-PING-INTERVAL" <>
+        value 30 <>
+        help "The interval of time (in seconds) between two pings to the WebSocket clients. It is instrumental in keeping connections alive.")
 
   where
     environ var = foldMap value (lookup var environment)
--- a/server/src/Core.hs
+++ b/server/src/Core.hs
@@ -5,6 +5,7 @@ module Core
   EnqueueResult (..),
   Command (..),
   ServerState,
+  SubscriberState (..),
   Updated (..),
   enqueueCommand,
   tryEnqueueCommand,
@@ -12,6 +13,7 @@ module Core
   withCoreMetrics,
   lookup,
   newCore,
+  newSubscriberState,
   postQuit,
   runCommandLoop,
   runSyncTimer
@@ -19,7 +21,7 @@ module Core
 where
 
 import Control.Concurrent (threadDelay)
-import Control.Concurrent.MVar (MVar, newMVar, putMVar)
+import Control.Concurrent.MVar (MVar, newEmptyMVar, newMVar, putMVar)
 import Control.Concurrent.STM (atomically)
 import Control.Concurrent.STM.TBQueue (TBQueue, newTBQueueIO, readTBQueue, writeTBQueue, isFullTBQueue)
 import Control.Concurrent.STM.TVar (TVar, newTVarIO)
@@ -30,8 +32,6 @@ import Data.Foldable (forM_, for_)
 import Data.Traversable (for)
 import Data.UUID (UUID)
 import Prelude hiding (log, writeFile)
-
-import qualified Network.WebSockets as WS
 
 import Config (Config (..), periodicSyncingEnabled)
 import Logger (Logger)
@@ -66,18 +66,35 @@ data Core = Core
   -- the "dirty" flag is set to True whenever the core value has been modified
   -- and is reset to False when it is persisted.
   , coreValueIsDirty :: TVar Bool
-  , coreQueue :: TBQueue Command
-  , coreUpdates :: TBQueue (Maybe Updated)
-  , coreClients :: MVar ServerState
-  , coreLogger  :: Logger
-  , coreConfig  :: Config
-  , coreMetrics :: Maybe Metrics.IcepeakMetrics
+  , coreQueue        :: TBQueue Command
+  , coreUpdates      :: TBQueue (Maybe Updated)
+  , coreClients      :: MVar ServerState
+  , coreLogger       :: Logger
+  , coreConfig       :: Config
+  , coreMetrics      :: Maybe Metrics.IcepeakMetrics
   }
 
-type ServerState = SubscriptionTree UUID WS.Connection
+-- This structure describes the state associated to each subscriber in order
+-- to communicate with them, at the moment, a simple `MVar` as a communication
+-- channel. This is a newtype in order for it to be extensible without
+-- rewriting all call sites.
+newtype SubscriberState = SubscriberState
+  -- We don't need actual queues for subscribers because the only relevant value
+  -- for them is the last one. Since we have one producer and one reader, we can
+  -- rely on MVar as a simpler mechanism.
+  -- We can expect from the subscribers to receive all the updates in the 
+  -- absence of timeouts.
+  { subscriberData :: MVar Value }
+
+-- This structure keeps track of all subscribers. We use one SubscriberState per
+-- subscriber.
+type ServerState = SubscriptionTree UUID SubscriberState
 
 newServerState :: ServerState
 newServerState = empty
+
+newSubscriberState :: IO SubscriberState
+newSubscriberState = SubscriberState <$> newEmptyMVar
 
 -- | Try to initialize the core. This loads the database and sets up the internal data structures.
 newCore :: Config -> Logger -> Maybe Metrics.IcepeakMetrics -> IO (Either String Core)
--- a/server/src/Subscription.hs
+++ b/server/src/Subscription.hs
@@ -21,7 +21,6 @@ import Data.Hashable (Hashable)
 import Data.Maybe (fromMaybe)
 import Data.Text (Text)
 
-import qualified Control.Concurrent.Async as Async
 import qualified Data.HashMap.Strict as HashMap
 import qualified Data.Text as Text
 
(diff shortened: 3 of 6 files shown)


server/src/Core.hs
```diff
@@ -76,21 +74,20 @@ data Core = Core
   , coreMetrics      :: Maybe Metrics.IcepeakMetrics
   }
 
-data SubscriberState = SubscriberState
-  { subscriberConnection :: WS.Connection
-  , subscriberQueue      :: TBQueue Value
-  }
+-- This structure describes the state associated to each subscriber in order to
+-- communicate with them, at the moment, a simple bounded queue. This is a
```
rkrzr, line 78: I would explain why we need the bounded queue here (=> because every subscriber can respond at a a different speed, and we don't want to wait until the slowest subscriber has responded. And the queue is bounded because we don't want to have too many messages pile up, since that would be a DOS vector)

server/src/Subscription.hs
```diff
@@ -77,18 +76,16 @@ unsubscribe path subid (SubscriptionTree here inner) =
...
   where notifications = broadcast' path value tree
 
 -- Like broadcast, but return a list of notifications rather than invoking an
 -- effect on each of them.
-broadcast' :: [Text] -> Value -> SubscriptionTree id conn -> [(conn, Value)]
+broadcast' :: [Text] -> Value -> SubscriptionTree id state -> [(state, Value)]
 broadcast' = \path value tree -> execWriter $ loop path value tree
   where
-  loop :: [Text] -> Value -> SubscriptionTree id conn -> Writer [(conn, Value)] ()
+  loop :: [Text] -> Value -> SubscriptionTree id state -> Writer [(state, Value)] ()
```
fatho, line 88: I suppose we rarely have updates that affect a lot of subscribers at once, otherwise, this could also be an issue due to using `Writer`. For one, even the "strict" writer is still lazy in the value, so it's building large thunks, and the monoid used here (`[]`) has `O(n)` append.

Not an issue to be solved in this PR, but something to keep in mind.

server/src/WebsocketServer.hs
```diff
@@ -96,29 +87,58 @@ authorizePendingConnection core conn
 handleClient :: WS.Connection -> Path -> Core -> IO ()
 handleClient conn path core = do
   uuid <- newUUID
+  -- TODO: Add a config field to retrieve the length for this queue.
+  subscriberState <- newSubscriberState 100
```
fatho, line 91: With this TODO being fixed this looks like a viable option. I'd probably use a smaller value here though, we expect subscribers to be fast. If they aren't, they deserve losing the updates.

server/src/WebsocketServer.hs
```diff
@@ -40,22 +41,12 @@ newUUID = randomIO
...
-      -- Rethrow async exceptions, they are meant for inter-thread communication
-      -- (e.g. ThreadKilled) and we don't expect them at this point.
-      | Just asyncExc <- fromException exc = throwIO (asyncExc :: SomeAsyncException)
-      -- We want to catch all other errors in order to prevent them from
-      -- bubbling up and disrupting the broadcasts to other clients.
-      | otherwise = pure ()
+    writeToSubQueue :: TBQueue Value -> Value -> IO ()
+    writeToSubQueue queue val = atomically $ do
+      isFull <- isFullTBQueue queue
+      unless isFull $ writeTBQueue queue val
```
fatho, line 47: For a production ready version, it would be interesting to count how many updates we lose due to individual queues being full (so just having a counter metric here for the `isFull` case).
That allows us to measure whether the chosen queue length is right (just a few lost events occasionally) or too small (many events lost).

fatho (commented):
Looks like a good approach, just some polishing remarks below.

Also, I think there were more potential solution ideas floating around, did you collect them somewhere and wrote down their pros and cons? Would be good to document how we ended up with this new design.

server/src/Subscription.hs
rkrzr: Good point. We should keep track of this somewhere. @DiegoDiverio can you open an issue for this and just note it down as a potential future improvement. You can then just put a comment here in the code and link to that issue.

server/src/WebsocketServer.hs
rkrzr: I think it would also be good to log a message every time we discard a message, due to the queue being full.

ReinierMaas (commented):
I have an open question about the design:

Should we drop the message that we try to insert?

Another option is to drop the next message that we are about to send.

---

A simple representation to get the idea across:

```
queue (max 3): m3->m2->m1
update: m4
choice drop m1 or drop m4:
* queue (drop m1): m4->m3->m2
* queue (drop m4): m3->m2->m1
```

Pros:

We will eventually get the client to the last known state

Cons:

We will skip intermediate states from the client's point of view

server/src/Subscription.hs
alex-mckenna: More recent `transformers` / `mtl` have a CPSd version of `Writer` which is less space-leaky. Maybe that's worth using since it's just an import change?

DiegoDiverio:
> Another option is to drop the next message that we are about to send.

This is a very good idea, I think. Because if you want to have something like a progressive evolution while ensuring you don't lose any of the information, I don't think you should use a KV store like Icepeak, since it doesn't provide this kind of certainty.

But, conversely, if you're only interested in the current/last value for a given key, a KV store is a good fit.

server/src/Subscription.hs
DiegoDiverio: If you're referring to https://hackage.haskell.org/package/mtl-2.3.1/docs/Control-Monad-Writer-CPS.html, I think that's a good idea, we just need a version bump for `mtl` (we have `< 2.3` as a version restriction at the moment) and the import change.

server/src/WebsocketServer.hs
rkrzr: > What extra information would a log give us in this case that we wouldn't get from a counter metric?

I would log the actual message that is being discarded (which you wouldn't get from the counter). I think this is useful for debugging, since you would get a direct log message and otherwise you would need to know that IP discards messages sometimes and would need to check the counter metric.

> I would expect such a log to be rather spammy.

I think it shouldn't be too bad, since we should just close the TCP connection to any customer that is timing out (@DiegoDiverio is that what happens now btw? Is the timeout value configurable?). So you would only get some log spam until the connection is closed.

DiegoDiverio:
A summary of the improvements to be included in this PR:
- Discarding the oldest element instead of the newest one.
- Add configuration options to modify the length of the subscriber queues and the duration of the timeout.

Not included in this PR:
- Change import and `mtl` version to use CPS `Writer` monad.
- Add logs and metrics to have information on Icepeak's behaviour regarding subscribers.

fatho:
> By the way, in the case the queue gets full, why not just empty the queue, since the intermediate values will be out-of-date? What is the use case for sending them all to the subscriber? We might just even need to have MVar (or another appropriate mutable state) instead of queues, and just register if the value has been updated.

Makes sense, a subscription always gets the full values for the path it registered, right? Then indeed only the most recent change would matter for the existing uses.

rkrzr:
> This is true, but we should use an explicit structure that conveys our intent, whichever one it is. I think MVar might convey this better, but a 1-element queue is fine as well, as long as we structure the code so that the intention is clear.

Good idea. For all of our current use cases an MVar would work just fine.

(I was first thinking that there might be other cases where intermediate states are important, say, a state machine, but since that's not a current requirement, let's not optimize for that)

server/src/WebsocketServer.hs
fatho: > I would log the actual message that is being discarded (which you wouldn't get from the counter). I think this is useful for debugging, since you would get a direct log message and otherwise you would need to know that IP discards messages sometimes and would need to check the counter metric.

I don't think that would help with debugging, as a client being too slow to receive data (or just a connection that was silently dropped) is not a problem we can reasonably take action on.

----

Adding a metric does have value though, because it allows us to actually observe that things are working correctly. (Logs would to of course, but a metric is both easier to observe and has lower overhead). That's why I would also disagree with

> Though I would advise against getting this PR too large: we should solve a problem at a time to keep the PR readable

The change in this PR is pretty big in terms of how icepeak operates, and leaves us with no way of observing that it is working correctly, and not losing half the updates or something. (We would probably eventually notice if it affects what our users see - but that's too late)

A simple counter metric would be a low overhead way of achieving this: as long as the metric is zero (or rising very slowly) it means that the updates are still working as intended.

If you want to keep the PR small, you can create a second PR to be based on this one, then merge them together before deploying.

rkrzr: Ok, makes sense. Let's go with the counter metric.

DiegoDiverio: Yeah, that's exactly the dev pattern I was looking for. I won't deploy before having merged all the developments we talked about here.

server/src/WebsocketServer.hs
```diff
@@ -73,11 +69,12 @@ acceptConnection core pending = do
         }
     AuthAccepted -> do
       let path = fst $ Uri.decodePath $ WS.requestPath $ WS.pendingRequest pending
+          timeout = configSubscriberTimeout $ coreConfig core
       connection <- WS.acceptRequest pending
       -- Fork a pinging thread, for each client, to keep idle connections open and to detect
       -- closed connections. Sends a ping message every 30 seconds.
       -- Note: The thread dies silently if the connection crashes or is closed.
-      WS.withPingThread connection 30 (pure ()) $ handleClient connection path core
+      WS.withPingThread connection timeout (pure ()) $ handleClient connection path core
```
fatho, line 77: The above comment (and subsequently looking into the implementation) makes me think that the 30 isn't actually a timeout. It seems to only control how often the server sends a keep-alive ping to prevent intermediate proxies from closing the connection.

In our case, such a (reverse) proxy is nginx which we use for the public-facing side of icepeak to terminate HTTPS connections. it seems that the actual timing out part is then happening when nginx hasn't heard from the client in a while and then closes the connection. That connection close is then received by icepeak, which then in turn makes the send operation in the ping thread, and in the update thread, and the receive operation in `keepTalking` fail.
(NB: Since for us, nginx and icepeak run on the same host at the moment, Nginx closing the connection will always be observed as there's no packet loss on a localhost connection - or shouldn't be at least). If those were run on different servers, the packet signalling that the connection was closed might actually get lost, resulting in a truly stuck connection until the kernel decides the connection is dead - which AFAIK is a way longer timeout).

It's still a good idea to make this configurable since other setups might need other ping intervals, but it's not doing what we intend to do. Adding proper timeouts within icepeak is a different issue.

server/src/Config.hs
```diff
@@ -105,6 +107,11 @@ configParser environment = Config
   <*> storageBackend
   <*> switch (long "sync-log" <>
              help "Enable logging the duration of Sync operations.")
+  <*> option auto
+       (long "subscriber-timeout" <>
```
fatho, line 111: This should rather be called `websocket-ping-interval`, see my other comment.

fatho (changes requested):
The MVar based queue looks good, just a remark about the "timeout":

server/src/WebsocketServer.hs
DiegoDiverio: Thanks for the explanation! I wasn't sure about what closed the connection. I'll create an issue to include timeout in Icepeak itself. Do you think that asking for a Pong using [ServerOptions.serverRequirePong](https://hackage.haskell.org/package/websockets-0.12.7.3/docs/Network-WebSockets.html#t:ServerOptions) could fit the bill?

fatho: A pong should always be sent upon receiving a ping, according to the websocket protocol. So that option seems indeed like what we would need: if a pong wasn't received for the specified interval, the connection is closed.

server/src/Config.hs
```diff
@@ -105,6 +107,11 @@ configParser environment = Config
   <*> storageBackend
   <*> switch (long "sync-log" <>
              help "Enable logging the duration of Sync operations.")
+  <*> option auto
+       (long "websocket-ping-interval" <>
+        metavar "WS-PING-INTERVAL" <>
+        value 30 <>
+        help "The interval of time between two pings to the WebSocket clients. It is instrumental in keeping connexions alive.")
```
fatho, line 114: Somewhere this should mention that this value is specified in seconds. (Could just use `metavar "SECONDS"` - then it shows up as `--websocket-ping-interval SECONDS` in the help text)

Also there's a small typo: "connections" rather than "connexions".

fatho (commented):
Small nit regarding the renamed option:

fatho (approved):
LGTM!

DiegoDiverio:
@OpsBotPrime merge

OpsBotPrime:
Rebased as 13b214b1ae5c8b5ce81d8066e6c16417362950bd, waiting for CI …

Issue #103, opened 2023-03-29
Expose metrics on subscribers


DiegoDiverio:
On top of https://github.com/channable/icepeak/pull/101, we should have some metrics to monitor the behaviour of subscribers, like the number of discarded messages.

PR #106, opened 2023-03-30
Increase WebSocket skipped updates metric when queue is not fetched

DiegoDiverio:
Fixes https://github.com/channable/icepeak/issues/103.

I reused the already existing metric, because the user-facing issue would be the same, and we can easily know in which case we are looking at the size of the queue (which can be computed from other metrics).

### DIFF
--- a/server/src/Metrics.hs
+++ b/server/src/Metrics.hs
@@ -19,21 +19,22 @@ type HttpStatusCode = Text
 type HttpRequestHistogram = Vector (HttpMethodLabel, HttpStatusCode) Histogram
 
 data IcepeakMetrics = IcepeakMetrics
-  { icepeakMetricsRequestCounter    :: HttpRequestHistogram
+  { icepeakMetricsRequestCounter        :: HttpRequestHistogram
   -- TODO: the following line can be removed after dashboard has been updated to use icepeak_data_size_bytes
-  , icepeakMetricsDataSize          :: Gauge
-  , icepeakMetricsDataSizeBytes     :: Gauge
-  , icepeakMetricsJournalSize       :: Gauge
-  , icepeakMetricsDataWritten       :: Counter
-  , icepeakMetricsDataWrittenTotal  :: Counter
-  , icepeakMetricsJournalWritten    :: Counter
-  , icepeakMetricsSubscriberCount   :: Gauge
-  , icepeakMetricsQueueAdded        :: Counter
-  , icepeakMetricsQueueRemoved      :: Counter
-  , icepeakMetricsSyncDuration      :: Histogram
-  , icepeakMetricsWsQueueAdded      :: Counter
-  , icepeakMetricsWsQueueRemoved    :: Counter
-  , icepeakMetricsWsSkippedUpdates  :: Counter
+  , icepeakMetricsDataSize              :: Gauge
+  , icepeakMetricsDataSizeBytes         :: Gauge
+  , icepeakMetricsJournalSize           :: Gauge
+  , icepeakMetricsDataWritten           :: Counter
+  , icepeakMetricsDataWrittenTotal      :: Counter
+  , icepeakMetricsJournalWritten        :: Counter
+  , icepeakMetricsSubscriberCount       :: Gauge
+  , icepeakMetricsQueueAdded            :: Counter
+  , icepeakMetricsQueueRemoved          :: Counter
+  , icepeakMetricsSyncDuration          :: Histogram
+  , icepeakMetricsWsQueueAdded          :: Counter
+  , icepeakMetricsWsQueueRemoved        :: Counter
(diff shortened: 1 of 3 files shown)


fatho (commented):
I think it would be better to have a different metric for this, as ultimately we're counting different things here:
- At this point, one count means one client didn't receive one update, and the cause can be narrowed down to the sending part.
- For the other queue, one count means *all* clients didn't receive one update, and the cause is likely to be found elsewhere.

It's always easy to make an aggregate dashboard that shows both (or just the sum of both), but getting the individual information from a compound metric after the fact is often hard.

DiegoDiverio:
I think I see your point. Though, a bit of detail: when we discard an update in the main update queue (which shouldn't ever happen, now), we don't know which client is affected, but only one of them will be, under our current use, and we're sure only one path will be affected.

However, I agree that it would make the information easier to see if we have two different metrics.

fatho:
> we don't know which client is affected, but only one of them will be

There still can be multiple clients subscribed to the same path, so while indeed it likely doesn't affect all of them if we lose one update to the `coreUpdates` queue, it's also not necessarily limited to a single client. There could also be no client at all listening on that path. All in all, that makes it even harder to assess the impact of a skipped update at that point.

server/src/WebsocketServer.hs
```diff
@@ -34,21 +34,25 @@ import JwtMiddleware (AuthResult (..), isRequestAuthorized, errorResponseBody)
...
     writeToSub queue val = do
       -- We are the only producer, so either the subscriber already
       -- read the value or we can discard it to replace it with the
       -- new one. We don't need atomicity for this operation.
       -- `tryTakeMVar` basically empties the MVar, from this perspective.
-      _ <- tryTakeMVar queue
+      mbQueue <- tryTakeMVar queue
+      -- If the MVar has not yet been read by the subscriber thread, it means
+      -- that the update has been skipped.
+      when (isJust mbQueue) $ for_ (coreMetrics core) Metrics.incrementWsSkippedUpdates
```
fatho, line 55: Shouldn't this now use the new `incrementWsQueueSkippedUpdates` function?

server/src/Metrics.hs
```diff
@@ -61,6 +62,8 @@ createAndRegisterIcepeakMetrics = IcepeakMetrics
                               "Total number of items added to the WebSocket queue."))
   <*> register (counter (Info "icepeak_internal_ws_queue_items_removed"
                               "Total number of items removed from the WebSocket queue."))
+  <*> register (counter (Info "icepeak_internal_ws_queue_skipped_updates"
+                              "Total number of updates discarded from the WebSocket queue."))
   <*> register (counter (Info "icepeak_internal_ws_skipped_updates_total"
                               "Total number of updates that have not been sent to subscribers."))
```
fatho, line 68: Ah, I see your commit message mentions "rename old metric". Given that prometheus doesn't support renames, those are always pretty disruptive since they break existing dashboards/severely complicate their queries (until the old data is completely flushed out).

If possible, it's better to find a new descriptive name for the new metric, and maybe adjust the help texts to help disambiguate the two.

DiegoDiverio: That is right, but to be honest, I was thinking about flushing the old metric. I can try and find better less disruptive names, though.

server/src/Metrics.hs
```diff
@@ -122,5 +125,8 @@ incrementWsQueueAdded = incCounter . icepeakMetricsWsQueueAdded
 incrementWsQueueRemoved :: MonadMonitor m => IcepeakMetrics -> m ()
 incrementWsQueueRemoved = incCounter . icepeakMetricsWsQueueRemoved
 
+incrementWsQueueSkippedUpdates :: MonadMonitor m => IcepeakMetrics -> m ()
+incrementWsQueueSkippedUpdates = incCounter . icepeakMetricsWsQueueSkippedUpdates
```
fatho, line 129: This function appears unused - I think this is the one you meant to use for the other queue now, given you resolved my other comment about this?

fatho (changes requested):

fatho (approved):
:shipit:

DiegoDiverio:
@OpsBotPrime merge

OpsBotPrime:
Pull request approved for merge by @DiegoDiverio, rebasing now.

OpsBotPrime:
Rebased as 65259c12c08c6c536bc00e33d67de113766faf89, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/c50fa957-a73c-4de5-9918-aa54e9b0b179?pipeline_id=56b96b58-fa33-43cc-b446-794a0f0e2680) started.

PR #112, opened 2023-05-09
Avoid O(n^2) time broadcast' by replacing Writer

### LINKED ISSUE #102: Update `Writer` to the CPS version in `mtl-2.3.1`
The `Writer` monad we use in `WebsocketServer.hs` might prove leaky. Following https://github.com/channable/icepeak/pull/101#discussion_r1150257486, we could switch to the CPS version.

However, this would require modifying a lot of versions for the dependencies, as well as using packages outside of Stackage: there is no Stackage package set that contains `mtl-2.3` or after. We could otherwise pin very precise versions in `extra-deps` and remove the version constraints in `package.yaml` and only pin the GHC version in `stack.yaml` as a resolver. 

I tried to play a bit with that, but as far as it's not in any Stackage set, it will call for `extra-deps`. So far, I ended up with what `stack` sees as a cycle, but actually is a version mismatch it can't solve.

I could get something working by flooding `extra-deps` and changing the relevant parts of the code that were breaking, but that's not a very nice solution.

robbert-vdh:
As mentioned by @fatho in #102, the Writer monad can be replaced with the strict state monad to avoid the use for mappend/list concatenation for every element added to the list.. This brings the time complexity down from `O(n^2)`  to `O(n)`. I added the reverse so the order stays the same before and after this change. If that's not needed the reverse can also be removed. Then order of the lists in the tests would then need to be changed.

Resolves #102.

### DIFF
--- a/server/src/Icepeak/Server/Subscription.hs
+++ b/server/src/Icepeak/Server/Subscription.hs
@@ -13,7 +13,7 @@ module Icepeak.Server.Subscription
 where
 
 import Control.Monad (void)
-import Control.Monad.Writer (Writer, tell, execWriter)
+import Control.Monad.State.Strict (State, execState, modify')
 import Data.Aeson (Value)
 import Data.Foldable (for_, traverse_)
 import Data.HashMap.Strict (HashMap)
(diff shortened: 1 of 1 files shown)


DiegoDiverio:
The code looks ok to me but how did you test it?

DiegoDiverio (approved):
LGTM! :rocket:

robbert-vdh:
@OpsBotPrime merge

OpsBotPrime:

PR #113, opened 2023-05-11
Add WebSocket time-out handling using the ping mechanism

### LINKED ISSUE #104: Add timeout handling to the WebSocket subscriber connections
As discussed in https://github.com/channable/icepeak/pull/101, we don't actually handle timeouts for WebSocket connections on Icepeak's side at the moment. It could be remedied by checking if the subscriber actually responds with a pong when we send a ping.

robbert-vdh:
As mentioned in #104, the while the websockets library does have a function that spawns a thread to periodically ping clients to keep the connection alive and to make sure the client hasn't timed out, it doesn't actually implement the latter part on its own (see https://github.com/jaspervdj/websockets/issues/159).

This implementation works by detecting timeouts, and then cancelling the websocket `app` action when that happens. The timeout detection is implemented by making sure the long pong was received within a certain time frame of the last sent ping whenever a new ping is sent. When that happens, the ping handler thread exits which causes the app action to exit as well thanks to the use of `race_` in `withInteruptablePingThread`. The original library-provided `withPingThread` runs the ping handler on its own thread with no clean way for the ping handler to affect the thread the `app` action is running on due to the use of `withAsync` (but without actually using the async value at any point).

The last piece of the puzzle is the pong handler, which is set through the `ConnectionOptions`. These options were previously shared for all connections, so a small refactor had to be made to allow different options for each connection.

Let me know if there are any questions or if anyone has suggestions on how to improve this. I found the easiest way to test this to be by launching the server with short timeouts (e.g. `stack run -- --websocket-ping-interval=1 --websocket-pong-timeout=1`), and then using websocat to connect to the server (`websocat ws://localhost:3000 -vvvv --print-ping-rtts`), sending `SIGSTOP`/pressing Ctrl+Z to simulate a timeout.

Fixes  #104.

### DIFF
--- a/server/src/Icepeak/Server/Config.hs
+++ b/server/src/Icepeak/Server/Config.hs
@@ -47,6 +47,11 @@ data Config = Config
   , configSyncLogging :: Bool
   -- | Delay between two pings to the client (useful in order to keep the connexion alive)
   , configWebSocketPingInterval:: Int
+  -- | The timespan in seconds after sending a ping during which the client has
+  -- to respond with a pong. If no pong is sent within this timeout threshold
+  -- then the connection is considered to have timed out and it will be
+  -- terminated.
+  , configWebSocketPongTimeout:: Int
   }
 
 data MetricsConfig = MetricsConfig
@@ -112,6 +117,11 @@ configParser environment = Config
         metavar "WS-PING-INTERVAL" <>
         value 30 <>
         help "The interval of time (in seconds) between two pings to the WebSocket clients. It is instrumental in keeping connections alive.")
+  <*> option auto
+       (long "websocket-pong-timeout" <>
+        metavar "WS-PONG-TIMEOUT" <>
+        value 30 <>
+        help "The timespan in seconds after sending a ping during which the client has to respond with a pong. If no pong is sent within this timeout threshold then the connection is considered to have timed out and it will be terminated.")
 
   where
     environ var = foldMap value (lookup var environment)
--- a/server/src/Icepeak/Server/WebsocketServer.hs
+++ b/server/src/Icepeak/Server/WebsocketServer.hs
@@ -1,30 +1,39 @@
 {-# LANGUAGE OverloadedStrings #-}
 
 module Icepeak.Server.WebsocketServer (
+  WSServerApp,
+  -- The constructor is intentionally not exposed since the 'TimeSpec' should be
+  -- initialized using the monotonic clock
+  WSServerOptions(),
+  wsConnectionOpts,
+  mkWSServerOptions,
+
   ServerState,
   acceptConnection,
   processUpdates
 ) where
 
-import Control.Concurrent (modifyMVar_, readMVar)
-import Control.Concurrent.Async (withAsync)
-import Control.Concurrent.MVar (MVar, tryTakeMVar, putMVar, takeMVar)
+import Control.Concurrent (modifyMVar_, readMVar, threadDelay)
+import Control.Concurrent.Async (withAsync, race_)
+import Control.Concurrent.MVar (MVar, newMVar, putMVar, swapMVar, takeMVar, tryTakeMVar)
 import Control.Concurrent.STM (atomically)
 import Control.Concurrent.STM.TBQueue (readTBQueue)
-import Control.Exception (SomeAsyncException, SomeException, finally, fromException, catch, throwIO)
-import Control.Monad (forever, when)
+import Control.Exception (SomeAsyncException, SomeException, finally, fromException, catch, throwIO, AsyncException, handle)
+import Control.Monad (forever, when, void, unless)
 import Data.Aeson (Value)
 import Data.Foldable (for_)
 import Data.Text (Text)
 import Data.UUID
+import System.Clock (Clock (Monotonic), TimeSpec(..), getTime)
 import System.Random (randomIO)
 
 import qualified Data.Aeson as Aeson
 import qualified Data.ByteString.Lazy as LBS
+import qualified Data.Text as T
 import qualified Data.Time.Clock.POSIX as Clock
-import qualified Network.WebSockets as WS
 import qualified Network.HTTP.Types.Header as HttpHeader
 import qualified Network.HTTP.Types.URI as Uri
+import qualified Network.WebSockets as WS
 
 import Icepeak.Server.Config (Config (..))
 import Icepeak.Server.Core (Core (..), ServerState, SubscriberState (..), Updated (..), getCurrentValue, withCoreMetrics, newSubscriberState)
@@ -35,6 +44,43 @@ import Icepeak.Server.JwtMiddleware (AuthResult (..), isRequestAuthorized, error
 import qualified Icepeak.Server.Metrics as Metrics
 import qualified Icepeak.Server.Subscription as Subscription
 import Data.Maybe (isJust)
+
+-- | 'WS.ServerApp' parameterized over the last received pong timestamp. See
+-- 'WSServerOptions'.
+type WSServerApp = WSServerOptions -> WS.ServerApp
+
+-- | Any additional payload information information used by the websocket
+-- application.
+--
+-- Currently this is only used to keep track of the last received pong. This
+-- value is initialized to the current time when starting the application, and
+-- it is updated in a connection-specific @connectionOnPong@ handler. When a
+-- ping is sent, the IO action in 'withInteruptablePingThread' checks whether
+-- the client answered our last ping with a pong. If it hasn't, then the server
+-- will terminate the websocket connection as timeouts would otherwise leave
+-- zombie connections.
+newtype WSServerOptions = WSServerOptions { wsOptionLastPongTime :: MVar TimeSpec }
+
+-- | Builds the /per connection/-connection settings. This hooks up the
+-- connection's pong handler to the last received pong time MVar so timeouts can
+-- be detected in the ping thread's handlers.
+wsConnectionOpts :: WSServerOptions -> WS.ConnectionOptions
+wsConnectionOpts wsOptions =
+  WS.defaultConnectionOptions
+    { WS.connectionOnPong = pongHandler wsOptions
+    }
+
+-- | Initialize a 'mkWSServerOptions' so it can be used for timeout tracking.
+-- See 'WSServerOptions' for more information.
+mkWSServerOptions :: IO WSServerOptions
+mkWSServerOptions = do
+  -- See 'WSServerOptions' for more information, but this is used to keep track
+  -- of when the last pong was received so the websocket connection can be
+  -- terminated if the client stops sending them. It's initialized with the
+  -- current time so this also works as expected if the client never sends a
+  -- single pong.
+  lastPongTime <- getTime Monotonic
+  WSServerOptions <$> newMVar lastPongTime
 
 newUUID :: IO UUID
 newUUID = randomIO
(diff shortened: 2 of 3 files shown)


server/src/Icepeak/Server/WebsocketServer.hs
```diff
@@ -173,3 +221,67 @@ processUpdates core = go
...
+  return $! lastPong < lastPongDeadline
+
+-- | Similar to 'WS.withPingThread', except that it uses a combination of
+-- 'pingHandler' and the 'pongHandler' set in the websocket connection's pong
+-- handler to detect that the thread client stopped sending pongs. If that
+-- happens the @app@ action will be canceled immediately.
+--
+-- The @pingAction@ is exected on every ping, and it should return @True@ if the
+-- client has timed out and the connection should be terminated.
+withInteruptablePingThread :: WS.Connection -> Int -> IO Bool -> IO () -> IO ()
```
DiegoDiverio, line 260: Is should be `Interruptible` or `Interruptable`. The first form is the most common though, I'd go for that one.

server/src/Icepeak/Server/Config.hs
```diff
@@ -112,6 +117,11 @@ configParser environment = Config
         metavar "WS-PING-INTERVAL" <>
         value 30 <>
         help "The interval of time (in seconds) between two pings to the WebSocket clients. It is instrumental in keeping connections alive.")
+  <*> option auto
+       (long "websocket-pong-timeout" <>
+        metavar "WS-PONG-TIMEOUT" <>
+        value 30 <>
+        help "The timespan in seconds after sending a ping during which the client has to respond with a pong. If no pong is sent within this timeout threshold then the connection is considered to have timed out and it will be terminated.")
```
DiegoDiverio, line 124:
```suggestion
        help "The timespan in seconds after sending a ping during which the client has to respond with a pong. If no pong is sent within this timeframe then the connection is considered to have timed out and it will be terminated.")
```

DiegoDiverio (changes requested):
The code looks nice! Suggestions are mostly about typos.

As discussed offline, we can add some toggling mechanism, in case you don't need these extra threads.

robbert-vdh:
> As discussed offline, we can add some toggling mechanism, in case you don't need these extra threads.

I just checked, and right now connections are actually leaked in production. If you connect to the production icepeak server that's reverse proxied through nginx and then just stop responding to pongs without closing the connection the connection will remain active and leak resources. I tested this by connecting with websocat, suspending the process (SIGSTOP/Ctrl+Z), waiting a couple minutes, then disconnecting from internet, hibernating my machine, resuming from hibernate, reconnecting to the internet, and then resuming the websocat process. The connection was still alive and it was like nothing happened. So I'm thinking it's probably a good idea to enable this ping-pong timeout mechanism in production, but maybe we can just set the timeout to like five or ten minutes instead of 30 seconds. Then the odds of it terminating connections unexpectedly are lower but it will still prevent resource leaks. I'd need to check if feedscrubber automatically reconnects closed websocket connections though.

robbert-vdh:
Thinking about this some more, if you change the pong timeout to be 15 minutes or more then the current behavior shouldn't change but it will still terminate the connections if the nginx TCP ping timeout mechanism somehow flukes out.

server/src/Icepeak/Server/WebsocketServer.hs
```diff
@@ -173,3 +221,67 @@ processUpdates core = go
...
+-- terminate as well.
+interruptiblePingThread :: WS.Connection -> Int -> IO Bool -> IO ()
+interruptiblePingThread conn pingInterval pingAction
+  | pingInterval <= 0 = return ()
+  | otherwise = ignore `handle` go 1
+  where
+    go :: Int -> IO ()
+    go i = do
+      threadDelay (pingInterval * 1000 * 1000)
+      WS.sendPing conn (T.pack $ show i)
```
fatho, line 276: Couldn't we now theoretically get stuck on this one (due to the lack of timeouts) and thus never reach the place where it is supposed to exit?
Concretely: if the send buffer is already full because the other side is not responding, then this would likely block (though I didn't check) - and the default kernel timeouts for TCP connections are rather long IIRC.

IIRC one of the designs I saw used an external thread with a timer that's reset when receiving pongs, and killing the handler thread when it times out, that would not be prone to this.
(Though it also depends on how `sendPing` is implemented)

server/src/Icepeak/Server/WebsocketServer.hs
```diff
@@ -36,6 +45,43 @@ import qualified Icepeak.Server.Metrics as Metrics
...
+-- application.
+--
+-- Currently this is only used to keep track of the last received pong. This
+-- value is initialized to the current time when starting the application, and
+-- it is updated in a connection-specific @connectionOnPong@ handler. When a
+-- ping is sent, the IO action in 'withInterruptiblePingThread' checks whether
+-- the client answered our last ping with a pong. If it hasn't, then the server
+-- will terminate the websocket connection as timeouts would otherwise leave
+-- zombie connections.
+newtype WSServerOptions = WSServerOptions { wsOptionLastPongTime :: MVar TimeSpec }
```
fatho, line 62: I don't see any use of the synchronization part of the MVar, wouldn't a regular `IORef` with atomic updates suffice here?

fatho (commented):
Nice doc comments! Two remarks from a cursory glance:

server/src/Icepeak/Server/WebsocketServer.hs
robbert-vdh: I wasn't entirely sure whether the pong would be received on the same thread the ping was sent from, which is why I went with `MVar` to avoid reading partial writes. `TimeSpec` consists of two `Int64`s, so they can never be read or written to atomically without external synchronization.

server/src/Icepeak/Server/WebsocketServer.hs
robbert-vdh: I'd need to check if it uses a blocking `send()` or not, I also have no idea. A quick glance indicates that it uses this function, which sounds like it would block when the send buffer is full: https://hackage.haskell.org/package/network-3.1.2.9/docs/Network-Socket-ByteString.html

At worst it would behave the same as before, since the `race_` used to spawn the action should prevent the ping thread from outliving the socket as the `app` action will terminate when the socket is closed, causing the other action to be cancelled. But I'll do some digging to figure out what would happen. If the send buffer indeed fills up (though with pings being only a couple bytes long, that would take forever), we could add another `MVar` for the last ping time and then send the pings form a third thread so the current ping handler thread only has to compare the two `MVar`s.

robbert-vdh: I just checked, with a default Ubuntu install it can send `342855` pings before filling the send buffer and blocking. This number will decrease slightly as time goes on (because the ping message becomes larger; it sends a sequence number as a string) but not in a significant way.

This would be fine in practice, but moving the ping sending to another thread and then comparing the values form the `MVar`s in the `interruptiblePingThread` thread does not have this limitation. What do you suggest, go with the third thread just in case or accept that with a 30 second ping interval a 20 year timeout window is probably sufficiently large?

DiegoDiverio: I would go with less over-engineering, since we should never run into these limitations.

server/src/Icepeak/Server/WebsocketServer.hs
fatho: > `TimeSpec` consists of two `Int64`s, so they can never be read or written to atomically without external synchronization.

You're thinking Rust but writing Haskell :slightly_smiling_face:   `IORef`s are always boxed, and thus always represented as a pointer which can be both read and written atomically without external synchronization.
For portable code, that should probably still happen through `atomicModifyIORef`.
Also see https://hackage.haskell.org/package/base-4.18.0.0/docs/Data-IORef.html#g:2

robbert-vdh:
I copied and adapted the ping thread structure from the websockets library to avoid changing behavior too much, but as @DiegoDiverio pointed out even when the timeout detection is disabled by setting the ping interval to a negative or zero value will still spawn a (lightweight) thread, which is then immediately exited. I made a small change to avoid spawning the thread when pings are disabled, and I also documented that a ping interval of 0 or less causes Icepeak to stop sending pings.

server/src/Icepeak/Server/WebsocketServer.hs
robbert-vdh: Yes I guess that's true. Adding a third thread would be a lot more robust in that situation.

server/src/Icepeak/Server/WebsocketServer.hs
robbert-vdh: Ah yeah good point, shouldn't make any any real difference performance wise but I'll switch this out for an `IORef` since partial updates aren't possible anyways. I changed it to use `IORef`s in c467874538e15340aca1e8a21c0ad7c6af26c93a.

server/src/Icepeak/Server/WebsocketServer.hs
fatho: Actually, I think it's not just hypothetical (though Diego might remember that better), but IIRC our initial symptoms of icepeak getting stuck were caused by a send being blocked, which in turn blocked the internal outgoing queue from making progress and eventually it ran full.

DiegoDiverio: Yeah, it was the issue, because we didn't decouple it enough. But I wonder if something like that could actually happen here. Maybe going safe is the way then.

robbert-vdh: I thought about this for some more and just adding a third thread may not be entirely correct since it would need to use its own delay/timer routine, which may slowly drift out of phase compared to the thread that actually sends the pings. Using an actual timer like you suggested would work, but we don't already depend on a package that can do this like [`timers`](https://hackage.haskell.org/package/timers-0.2.0.4/docs/Control-Concurrent-Timer.html). I checked the implementation and I don't think it's that much less expensive to restart a timer there than it would be to just slap a `timeout` onto the `WS.sendPing`, which works as expected. Any objections to just doing that instead? That prevents the ping from blocking indefinitely when the send buffer is full, and it interacts nicely with the existing timeout checking function.

robbert-vdh: > Doesn't the current implementation also run out of phase, because the ping thread runs in intervals of "ping interval" while the pongs are expected after the "pong timeout"?

There's a bit of play between `pingAction` being evaluated and the on-pong handler being evaluated but that should follow a normal distribution and not really increase or decrease over time. The phase drifting I was talking about would happen if you used another thread for checking the `IORef TimeSpec`s written by the ping- and pong handler instead of using an actual resettable timer. But I think the `timeout` here is an even nicer solution since it doesn't need to introduce any more moving parts.

> Less expensive in terms of what? timeout also uses a timer internally. It's certainly less complex though - but what would be the timeout for that operation then?

I set the timeout to the ping interval. Since if the socket blocks for that long, then the `pingHandler` will also consider the connection to have timed out so it works out nicely without any other changes.

fatho: Sounds like a plan!

robbert-vdh:
I'll merge this, and then we can set the pong timeout to match the TCP keepalive timeout when deploying. Should be Fine:tm:.

@OpsBotPrime merge

OpsBotPrime:

OpsBotPrime:
Rebased as b831c1b54f41bc664b4955642b0c8fc0768e359f, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/9eb042d2-6419-4a45-a0e8-488affc5297d?pipeline_id=b55d03ca-9eb8-42ac-85e8-2f11f1da685c) started.

PR #114, opened 2023-05-26
Add content-type header to PUT request in client

RileyApeldoorn:
Attach the appropriate "Content-Type" header when sending a PUT request from the client.

Found this bug because axum expects requests with a JSON body to have this header set to the appropriate value.


robbert-vdh:
Sorry for not getting to this earlier! LGTM. It's not needed for the `icepeak-server` implementation, but the header should still be there if you're sending JSON.

PR #116, opened 2023-06-09
Switch everything over to Cabal

robbert-vdh:
Just like in our other projects. I also took the opportunity to move the Nix builds overt to the more modern overlay style where the `icepeak` and `icepeak-client` packages are added to the Haskell packages as an overlay. They're then exposed in `build.py` so the CI can easily build both of them separately. I've made these changes:

- Instead of treating this like two separate projects with their own `stack.yaml`, there's now a single `default.nix` file and the `cabal.project` file in the repository's root now references both packages. That means you can run `cabal run <icepeak/icepeak-client/all>` from anywhere in the repo to build/run the packages.
- There's now a `.envrc` file for Nix integration with direnv, since Hoff and all our other repos also have this.
- I haven't fully modernized the Seamaphore build, but I did remove the explicit stack step since it's redundant with the `nix-build` and that invokes Cabal anyways.
- I wasn't quite sure what to do with the integration tests, which is why I marked this as a draft to avoid merging. As of right now, half of them fail. Looking at the git logs it looks like they haven't been touched for a couple years. @fatho may have been one of the last persons to work on them. I changed the shell script ones to invoke Cabal instead of Stack, but there are also a couple standalone Haskell files in there. Do we want to keep these? If so, they should probably be moved to a proper integration test binary instead of being standalone files.


FPtje:
Asking @HugoPeters1024 to take a look, because I'm rather busy. @HugoPeters1024 Do you have time to look at this? Thanks!

server/integration-tests/data-survives-restart.sh
```diff
@@ -1,12 +1,10 @@
 #!/usr/bin/env bash
 
-if [ ! -f "stack.yaml" ]; then
-    echo "Run test in server package directory"
-    exit 1
-fi
+cd "$(dirname "$0")"
+cd ..
```
HugoPeters1024, line 4: Thanks I hate it (but can't think of why this is objectively bad)

.semaphore/semaphore.yml
```diff
@@ -8,47 +8,7 @@ agent:
     os_image: "ubuntu2004"
 
 blocks:
-  - name: "Run checks"
```
HugoPeters1024, line 11: are we just getting rid of running tests in CI?

.semaphore/semaphore.yml
```diff
@@ -85,11 +45,11 @@ blocks:
       jobs:
         - name: "Build client"
           commands:
-            - "nix-build --no-out-link -E '(import ./nix/nixpkgs.nix {}).haskellPackages.callPackage ./client-haskell/client.nix {}'"
```
HugoPeters1024, line 88: could you also incorporate running `cabal-fmt` somewhere in the CI? (formatter for the cabal file)

server/integration-tests/auth_stress_test.hs
```diff
@@ -1,6 +1,9 @@
 #!/usr/bin/env stack
 -- stack --stack-yaml ../../client-haskell/stack.yaml runghc --package QuickCheck --package quickcheck-text
 
+-- FIXME: These scripts will either need to be dropped, or they should be moved
```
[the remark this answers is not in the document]
robbert-vdh: They probably haven't been used in years since they won't run correctly as is. @fatho was the last one who worked on this. We can probably also just leave them as is and fix them up once they're needed again.

server/integration-tests/data-survives-restart.sh
robbert-vdh:
```shell
cd "$(dirname "$(dirname "$0")")"
```
doesn't work if the script is invoked relatively from the same directory or through a couple other names. This works, and I've done it in the Hoff repo, but it's a bit uglier so just doing the two `cd`s is probably better:

```bash
cd "$(dirname "$(dirname "$(realname "$0")")")"
```

.semaphore/semaphore.yml
robbert-vdh: No, they're still run. From the OP:

> - I haven't fully modernized the Seamaphore build, but I did remove the explicit stack step since it's redundant with the `nix-build` and that invokes Cabal anyways.

.semaphore/semaphore.yml
robbert-vdh: In the other repos it's part of `build.py` but we don't have a `build.py` here. Or any other lints for that matter. The CI just checks if it builds and the tests pass. There are currently no lints or style checks at all.

HugoPeters1024 (approved):

server/integration-tests/auth_stress_test.hs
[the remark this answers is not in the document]
robbert-vdh: I tried this before but I now found the actual ticket that was closed with a #wontfix. I had already tried converting the scripts to cabal run scripts but that didn't work. This is why: https://github.com/haskell/cabal/issues/8562

robbert-vdh:
Since they currently don't run cleanly anyways, we decided to just look into fixing up the `stack runghc`-based integration later if we ever need to run them again.

@OpsBotPrime merge

OpsBotPrime:

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/36b2989c-358d-40dd-bd9a-1956a3e52c39?pipeline_id=02b7a218-5e83-4728-89ce-c07b694befc6) started.

PR #117, opened 2023-06-28
Freeze the package versions based on the ones from nixpkgs

robbert-vdh:
This prevents the plain Cabal and the Nix builds from using different versions. This is especially important when using Nix, as Cabal may start using more up to date versions of the package from Hackage if the user has run `cabal update`  before and a local package database exists. Cabal can also be forced to use only the global package registry in which case this problem also doesn't exist, but that requires either breaking non-Nix builds or wrapping the `cabal` command.

I also added the freeze files to the Nix builds so we'll get alerted if nixpkgs gets updated without also updating the freeze file.


robbert-vdh:
I changed the check to just use `cabal build --dry-run all` instead of making it part of the Nix build. Having it be part of the Nix build could cause problems if the derivation was included as part of a nixpkgs overlay with a different nixpkgs version.

.semaphore/semaphore.yml
```diff
@@ -43,6 +43,10 @@ blocks:
           - echo "max-jobs = auto" >> ~/.config/nix/nix.conf
 
       jobs:
+        - name: "Check frozen versions"
+          commands:
+            - "cabal build --dry-run all"
```
fatho, line 48: Might need a `nix develop` - it failed on CI because it cannot find cabal

robbert-vdh: You're right, I forgot there's no devenv there. Should be fixed now.

fatho (approved):
LGTM

OpsBotPrime:

OpsBotPrime:
Rebased as 4173a027d066c1f34bf74c99b0cb415feb037865, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/4ccfce80-602b-4a3d-9b3b-db464841ad78?pipeline_id=933b1907-8421-4b02-8014-6a5398328d63) started.

PR #125, opened 2024-12-20
Upgrade to the latest nixpkgs revision (via 'niv update') which comes with GHC 9.6.6

rkrzr:
Upgrade to the latest nixpkgs revision (via 'niv update') which comes with GHC 9.6.6

- Bump the nixpks pin
- Remove the upper version bounds in cabal in order to upgrade
- Update all libraries to the versions provided by Nix
- Fix some fallout (Scotty handles errors a bit differently now, some imports were moved around)
- Re-freeze cabal.project.freeze


ReinierMaas (approved):
LGTM! Thanks for the update.

PR #132, opened 2026-04-16
Add /healthcheck endpoint to the metrics server

rkrzr:
The metrics server was being misused as a Consul health check, since
icepeak had no dedicated liveness endpoint. This adds GET /healthcheck
on the same host/port as the metrics server (configured via --metrics),
returning 200 OK with an empty body.

The routing is handled inside a new metricsApp WAI application that
intercepts GET /healthcheck and falls through to the Prometheus metrics
app for all other requests. No new dependencies are required.


maartenberg (approved):
LGTM, thanks!
Will you also ship the new release? If so, please bump the version numbers in `server/icepeak.cabal` and `server/icepeak.nix`.

PR #134, opened 2026-09-16
Add timing metric for internal event handling

fatho:
This metric is essential for determining the remaining capacity of the single-threaded command loop in production.

There is now some overlap with the "queue removed" metric (but with the distinction that the new metric is split by event type, and only counted after handling is completed, so I think it's fine to keep both) and the "sync duration" metric, which we should eventually phase out. For now, I kept it though to not break any existing dashboards and alerts.


radekchannable (approved):
LGTM!

fatho:
@OpsBotPrime merge

OpsBotPrime:
Rebased as d51f32e56fbe6e8b2ec75ad254bb34f5803d2aeb, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/b225b4c8-b040-4a62-8246-e83322bdad6a?pipeline_id=946e6456-2260-453c-a668-85883c69a794) started.

PR #135, opened 2026-09-22
Fix missing strictness in updates

robbert-vdh:
We've been observing faster than expected growth in RTS live bytes in production. There don't seem to be that many places where we the Icepeak server can suffer traditional space leaks from thunk buildups, but one place where strictness could make a difference are the document state updates. In practice most of these state update thunks should have be evaluated quickly when the updates are sent to connected websocket clients, but if a client stops receiving messages then a thunk referencing the old state may stay alive until the client is terminated. Without measurements it's hard to tell what kind of difference this will make, but using a strict `modifyTVar'` here should at least help ensure updates don't build up thunks.


fatho (approved):
LGTM

robbert-vdh:
@OpsBotPrime merge

OpsBotPrime:
Rebased as df9dcf8134d7e1bd2b372ee44bd3f4e7888c0c50, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/7972e889-89c5-4a62-b3df-a1cbd1de9d2c?pipeline_id=205f1218-8efc-442c-8f4a-40d0159fc197) started.

PR #136, opened 2026-09-22
Reduce allocations in websocket broadcasting code

robbert-vdh:
Related to the investigation that prompted #135, but I split this off into its own PR since the changes are not really related.

While I was trying to find more low hanging fruit explanations for the space leak-like behavior we have been observing in production, I noticed `broadcast` was producing intermediate data structures that did not get optimized away (confirmed using `-ddump-simpl`. I'm not sure if this is going to make a measurable difference because I have not yet been able to produce the production memory behavior locally, but getting rid of the intermediate data structures also makes the code slightly simpler so it's probably doesn't hurt. 

I also made the updates of the subscriber tree strict while I was at it. In practice that should not never cause any issues, but clients rapidly connecting and disconnecting without any  updates being sent would still result in thunks unnecessarily building up.


server/src/Icepeak/Server/Subscription.hs
```diff
@@ -70,38 +68,23 @@ unsubscribe path subid (SubscriptionTree here inner) =
...
-        traverse_ (\v -> modify' ((v, value) :)) here
-        let broadcastInner key = loop [] (Store.lookupOrNull [key] value)
-        void $ HashMap.traverseWithKey broadcastInner inner
-
-      key : pathTail -> do
-        traverse_ (\v -> modify' ((v, value) :)) here
+      [] ->
+        traverse_
+          (\(key, subs) -> loop [] (Store.lookupOrNull [key] value) subs)
+          (HashMap.toList inner)
```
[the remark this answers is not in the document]
robbert-vdh: The intermediate list gets optimized away with `-O2`. The generated core is the same, and the `traverse_` version reads better.

server/src/Icepeak/Server/Subscription.hs
```diff
@@ -1,19 +1,17 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DeriveFunctor #-}
 
 module Icepeak.Server.Subscription
 (
   SubscriptionTree (..),
   broadcast,
```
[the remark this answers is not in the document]
robbert-vdh: `broadcast'` is no longer used or needed with this change.

server/src/Icepeak/Server/Core.hs
```diff
@@ -66,7 +67,7 @@ data Core = Core
     coreValueIsDirty :: TVar Bool
   , coreQueue :: TBQueue Command
   , coreUpdates :: TBQueue (Maybe Updated)
-  , coreClients :: MVar ServerState
+  , coreClients :: IORef ServerState
```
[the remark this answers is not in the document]
robbert-vdh: This part of the Icepeak server is not intended to be part of the public API.

fatho, line 70: What was the reason for switching the MVar to an IORef? Might be good to leave a comment about that.

server/src/Icepeak/Server/Subscription.hs
```diff
@@ -70,38 +68,23 @@ unsubscribe path subid (SubscriptionTree here inner) =
...
-  where
-  -- To prevent this from accidentally blowing up to @O(n^2)@ time, this uses
-  -- the strict state monad to cons the connections to a list, which then needs
-  -- to be reversed to get the correct order. This way the operation can be done
-  -- in @O(n)@ list operations where @n@ is the number of subscribers matched
-  -- along the path.
-  loop :: [Text] -> Value -> SubscriptionTree id state -> State [(state, Value)] ()
+broadcast f = loop
+ where
   loop path value (SubscriptionTree here inner) = do
```
fatho: Should `value` be a strict argument so that we never pass thunks of the store lookup to `f`?

fatho (commented):
Neat, just two small questions:

server/src/Icepeak/Server/Core.hs
robbert-vdh: The mutual exclusion part of the mutex was never used - all operations on it were either reads or pure modifications. IORef does the same but is slightly less footgun-y because it has a strict modification function. With `modifyMVar` you need to be careful that that you force evaluation, because `modifyMVar_ (pure . f)` ends up building large thunks until the value is read. This code was originally written in 2017, and we probably did not know as much about the tradeoffs between Haskell's various synchronization primitives at the time.

server/src/Icepeak/Server/Subscription.hs
robbert-vdh: Yeah good one! With the changes from the other PR `value` should always be in WHNF, but better safe than sorry.

fatho (approved):
:shipit:

robbert-vdh:
@OpsBotPrime merge

OpsBotPrime:
Pull request approved for merge by @robbert-vdh, waiting for rebase behind one pull request.

OpsBotPrime:
Speculatively rebased as e684238a7f8b6d92955c02527dc85bb37445ae87 behind 1 other PR, waiting for CI …

OpsBotPrime:
[CI job :yellow_circle:](https://channable.semaphoreci.com/workflows/abb87976-8042-46f3-b72d-f6ae1b653894?pipeline_id=1c56312f-cb53-4e6c-b70a-979bf1bfc2d5) started.

